PR #50, opened 2018-05-12
Use AssemblyLoadContext to customize native dependency load

tmat:
On Windows `DllImport` doesn't load a native dependency if it has already been loaded explicitly using `LoadLibrary`. That's not the case on Linux and the loader attempts to load the dependency from next to the referencing assembly.

To address this issue we implement a custom `AssemblyLoadContext` that handles the loading.


jkotas:
cc @jeffschwMSFT This PR is a good example of what one has to do to compensate for the lack of better plugin model. `GitLoaderContext` has the interesting code.

nguerrera:
@tmat Did the pre-loading work on some Linuxes but not others? I'm actually seeing P/Invoke to already loaded library work on Ubuntu, but I can't explain why based on this and other related issues.

PR #53, opened 2018-05-16
Correct property name

richlander:
I'm assuming that this was a typo.


tmat:
I believe the original is correct. If Deterministic property is true and ContinuousIntegrationBuild is true we set DeterministicSourcePaths  to true.

PR #57, opened 2018-05-22
Remove unnecessary net461 libs and implement AssemblyResolve event.

tmat:
AssemblyResolve is needed to load dependencies of netstandard.dll since their version 0.0.0.0 needs to be mapped to the actual assembly version.


tmat:
@bording The loading issues are never ending. This time there is an issue with all the facades of netstandard 2.0 that are needed when running on .NET 4.6.*. Would you mind multi-targeting LibGit2Sharp to `netstandard20` and `net461`? That would make withing msbuild tasks that use it much easier (and their packages smaller).

bording:
@tmat Is this because you don't have binding redirects that would handle this for you?

tmat:
@bording Exactly. Msbuild doesn't support binding redirects for tasks.

bording:
What scenario causes you to hit the problem you're fixing in this PR? I've been trying out the package and it's been working fine without this.

tmat:
@bording It repros on .NET 4.6.2.

bording:
FYI recent 0.26.0 previews of LibGit2Sharp are now multi-targeted to `net461` and `netstandard2.0`.

Issue #161, opened 2018-08-31
Integration in vcxproj not working


sepiel:
Hello, 

the integration of SourceLink inside vcxproj is not working. I added Microsoft.SourceLink.GitHub via NuGet to a project. The build is going well, but the pdb does not contain sourcelink information to the repository.

The issue can be seen in following test project:
[https://github.com/sepiel/Vcx.SourceLink.Test](https://github.com/sepiel/Vcx.SourceLink.Test)

PR #165, opened 2018-09-10
Add support for VC++

tmat:
Fixes https://github.com/dotnet/sourcelink/issues/161

### DIFF
--- a/src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
+++ b/src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
@@ -20,20 +20,33 @@
           BeforeTargets="CoreCompile"
           Condition="'$(EmbedUntrackedSources)' == 'true' and '$(EmbedAllSources)' != 'true' and '$(DebugType)' != 'none' and '$(EnableSourceControlManagerQueries)' == 'true'" />
 
-  <!-- 
-    If defined populates MappedPath metadata of SourceRoot items. 
+  <!--
+    If defined populates MappedPath metadata of SourceRoot items.
   -->
   <Target Name="_InitializeSourceRootMappedPathsOpt"
           DependsOnTargets="InitializeSourceRootMappedPaths"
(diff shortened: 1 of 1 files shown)


src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
```diff
@@ -42,6 +55,13 @@
     <ItemGroup>
       <FileWrites Include="$(SourceLink)" />
     </ItemGroup>
+
+    <!-- C++ Link task currently doesn't recognize SourceLink property -->
```
ctaggart: I think this block may solve https://stackoverflow.com/q/52423484/23059

quasarea: I'm doing it like this - defining custom cpp.props file:

    <?xml version="1.0" encoding="utf-8"?>
    <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
        <ItemDefinitionGroup>
            <Link>
                <GenerateDebugInformation>DebugFull</GenerateDebugInformation>
                <AdditionalOptions>/SOURCELINK:"${sourcelink}" %(AdditionalOptions)</AdditionalOptions>
            </Link>
        </ItemDefinitionGroup>
    </Project>
and then adding it to msbuild by /p:ForceImportBeforeCppTargets=cpp.props

will try your way now.

quasarea: Based on yours example I created custom target:

    <Target Name="IncludeSourceLinkFile" BeforeTargets="PreBuildEvent">
        <Message Text="IncludeSourceLinkFile..." />
        <ItemGroup Condition="'$(Language)' == 'C++'">
            <Link Update="@(Link)">
                <AdditionalOptions>%(Link.AdditionalOptions) /sourcelink:"$(SourceLink)"</AdditionalOptions>
            </Link>
        </ItemGroup>
    </Target>

and then added it via msbuild by /p:ForceImportBeforeCppTargets=cpp.props, it is executed but possibly too early as it does not affect link arguments

PR #191, opened 2018-10-17
Switch infrastructure to Arcade

tmat:


global.json
```diff
@@ -1,8 +1,8 @@
 {
-  "sdk": {
-    "version": "2.1.300"
+  "tools": {
+    "dotnet": "2.1.401"
   },
   "msbuild-sdks": {
-    "RoslynTools.RepoToolset": "1.0.0-beta2-63103-01"
+    "Microsoft.DotNet.Arcade.Sdk": "1.0.0-beta.18515.2"
```
ctaggart: What is Arcade?

eng/Versions.props
```diff
@@ -5,7 +5,7 @@
 
     <!-- This repo version -->
     <VersionPrefix>1.0.0</VersionPrefix>
-    <PreReleaseVersionLabel>beta</PreReleaseVersionLabel>
+    <PreReleaseVersionLabel>beta2</PreReleaseVersionLabel>
```
ctaggart: Is `PreReleaseVersionLabel` a new property?

global.json
tmat: https://github.com/dotnet/arcade/blob/master/Documentation/ArcadeSdk.md

tmat: Common infrastructure for .NET Core repos.

eng/Versions.props
tmat: It's property defined by Arcade.

PR #202, opened 2018-11-01
Add Arch Linux to RuntimeIdMap

akihikodaki:


bording:
I'm sure @tmat will point this out as well, but you can't just add arbitrary RIDs like this here. These need to be supported by .NET Core first. The files here are just an reflecting values that exist in the primary RID graph that comes from https://github.com/dotnet/corefx/tree/master/pkg/Microsoft.NETCore.Platforms.

akihikodaki:
Thanks for pointing out the source. I will close this issue.

Issue #241, opened 2019-02-07
SourceLink's usage of MSBuildAllProjects contributes significantly to VS memory consumption


davkean:
Starting in Dev16, all targets/props files should stop appending to `MSBuildAllProjects`, MSBuild will now automatically make sure that the first element of this property always contains the newest targets file, and hence consumption of it will continue to work.

Across a bunch of repos we've started removing setting of this property to reduce memory pressure, for example: https://github.com/dotnet/sdk/pull/2853.

This repo's targets files is contributing to significant memory usage when opening CoreFx. Using a [custom tool](https://github.com/drewnoakes/string-theory) for analyzing dumps for duplicate strings, ~12% of the heap (nearly 400 MB) is made up of the strings caused by the appending of this, which is SourceLink.

![image](https://user-images.githubusercontent.com/1103906/52389295-e8e9ec80-2ae6-11e9-8590-e6b198173aad.png)

PR #242, opened 2019-02-07
Remove MSBuildAllProjects

tmat:
Fixes https://github.com/dotnet/sourcelink/issues/241

### DIFF
--- a/eng/Signing.props
+++ b/eng/Signing.props
@@ -1,9 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
-  <PropertyGroup>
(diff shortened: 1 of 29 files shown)


bording:
These packages are still going to be working with older versions of MSBuild as well, right? If so, these might need to stay, but be conditional instead.

tmat:
They are not really needed even in Dev15. The only thing that would break is the incremental build in case you update the targets/props file in the nuget cache. I don't think we need to support that scenario as one should generally not be updating these files in nuget cache.

bording:
> I don't think we need to support that scenario as one should generally not be updating these files in nuget cache.

Yeah, good point.

Issue #258, opened 2019-04-06
Beta 2, Multiple Projects: Assembly with the same name is already loaded


lordmilko:
I have a solution consisting of two projects that both have sourcelink enabled.

Upon upgrading my these projects from `1.0.0-beta-63127-02` to `1.0.0-beta2-18618-05` however I am getting the following error when I try to `dotnet build` the solution

```
C:\Users\administrator\.nuget\packages\microsoft.build.tasks.git\1.0.0-beta2-18618-05\build\
Microsoft.Build.Tasks.Git.targets(20,5): error MSB4062: The "Microsoft.Build.Tasks.Git.LocateRepository" task
could not be loaded from the assembly C:\Users\administrator\.nuget\packages\microsoft.build.tasks.git
\1.0.0-beta2-18618-05\build\..\tools\netcoreapp2.0\Microsoft.Build.Tasks.Git.dll.
Assembly with same name is already loaded Confirm that the <UsingTask> declaration is correct, 
that the assembly and all its dependencies are available, and that the task contains a public class that implements
Microsoft.Build.Framework.ITask. 
[C:\Users\administrator\Desktop\Issue\Failure\Project2\Project2.csproj]
```

Attached are two solutions that help demonstrate this issue. 

The `Success` solution is using the older `1.0.0-beta-63127-02` version, while the `Failure` solution is using the newer `1.0.0-beta2-18618-05` version

I was able to replicate this issue on both my workstation running Windows 7 + .NET Core SDK 3.0.100-preview4-010483 and a Windows 10 workstation running .NET SDK 2.2.202

In the attached examples, I am using the default `PackageReference` definition that resulted from installing the package via the NuGet Package Manager, however I also seem to experience the issue using the more minimal definition described on the README

```xml
  <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta2-18618-05" PrivateAssets="All"/>
```

This issue appears to be different to #123 in that I am not trying to change anything between builds; I just want to do a single build!

I am also not sure whether this issue only happens when compiling from PowerShell/Core/ISE or cmd.exe also - I am getting mixed results

Neither solution in the attached zip file is linked with an actual repo or has any commits created, however that should not matter in terms of the underlying issue

## Reproducing

1. Download SourceLinkExamples.zip
2. Right click -> properties. Unblock file downloaded from internet (for good measure)
3. Under each subfolder run `dotnet build` via cmd.exe or PowerShell

[...]

PR #277, opened 2019-05-07
Auto-generate versions of build task assemblies

tmat:
Fixes https://github.com/dotnet/sourcelink/issues/258


global.json
```diff
@@ -1,8 +1,8 @@
 {
   "tools": {
-    "dotnet": "2.1.600-preview-009426"
+    "dotnet": "2.2.203"
```
ctaggart: Yay

PR #288, opened 2019-06-10
Initial implementation of direct git repo data reading

tmat:
Work in progress: replace usage of libgit2 APIs with fully managed implementation. This will make SourceLink available on all platforms that the Core CLR supports as well as in source build scenarios.


onovotny:
Is there a published nuget version that has these fixes in it?

tmat:
Not yet. Still testing the package.

Issue #332, opened 2019-07-09
SourceLink file should not regenerate json file if it doesn't need to be updated


jkotalik:
We are hitting an issue in ASP.NET Core which is causing unnecessary rebuilds of native dlls due to the .sourcelink.json file being updated.

Right now, the GeneratedSourceLinkFile task always creates the source link file, as shown here. https://github.com/dotnet/sourcelink/blob/c092238370e0437eb95722f28c79273244dc7f1a/src/SourceLink.Common/GenerateSourceLinkFile.cs#L115. This causes issues when invoking LINK with native projects, where it says source compilation is required with the error:
```
Source compilation required: input OUTOFPROCESSREQUESTHANDLER.SOURCELINK.JSON is newer than output ASPNETCOREV2_OUTOFPROCESS.DLL.
```

The simple fix is to check if the file content has changed or not between rebuilds.

PR #333, opened 2019-07-09
Check if file needs to be rebuilt before writing it

jkotalik:
Fixes https://github.com/dotnet/sourcelink/issues/332

If you'd like, I can make it so the test checks hashs instead of write time.

Also let me know if the target branch is correct.

### DIFF
--- a/src/SourceLink.Common.UnitTests/GenerateSourceLinkFileTests.cs
+++ b/src/SourceLink.Common.UnitTests/GenerateSourceLinkFileTests.cs
@@ -141,35 +141,29 @@ namespace Microsoft.SourceLink.Common.UnitTests
         [Fact]
         public void DoesNotRewriteContentIfFileContentIsSame()
         {
-            var tempFile = Path.GetTempFileName();
-            try
+            using var temp = new TempRoot();
+            var tempFile = temp.CreateFile();
+            
+            var engine = new MockEngine();
+            var task = new GenerateSourceLinkFile()
             {
-                var engine = new MockEngine();
-                var task = new GenerateSourceLinkFile()
+                BuildEngine = engine,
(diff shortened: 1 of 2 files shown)


jkotalik:
As far as I can tell, to get this update into arcade, I would need to update it here: https://github.com/dotnet/arcade/blob/1fb1e240c889cd7f6e10cb1eacd129efa3ddb4b4/src/Microsoft.DotNet.Arcade.Sdk/tools/DefaultVersions.props#L93

src/SourceLink.Common.UnitTests/GenerateSourceLinkFileTests.cs
```diff
@@ -137,5 +137,39 @@ public void Errors()
 
             Assert.Null(content);
         }
+
+        [Fact]
+        public void DoesNotRewriteContentIfFileContentIsSame()
+        {
+            var tempFile = Path.GetTempFileName();
```
tmat: Do not use `Path.GetTempFileName`. The method is unreliable.

Instead use:
```
using var temp = new TempRoot();
var tempFile = temp.CreateFile();
```

this will automatically delete it, so no need for `finally`

jkotalik: I'm looking at the docs and I don't really see why this API is unreliable: https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename?view=netframework-4.8. Mind clarifying why?

tmat: It fails after there is more than 65535 files temp created.

tmat:
Thanks for the fix!

PR #431, opened 2019-10-03
Source build fixes for target framework and project exclusions

crummel:
Two changes for source-build:
- Exclude test projects from source-build.
- Build only for .NET Core in source-build.  This allows us to get rid of the .NET Framework reference assembly prebuilt package.

cc @NikolaMilosavljevic @adaggarwal


crummel:
We were doing it to get rid of prebuilts, but we're trying out another approach that may mean we don't have to go to core-only, I'll update when we have more info.

crummel:
I think this is fixed between Aditya's work and the Arcade upgrade we took, closing this.

PR #447, opened 2019-10-15
Update BitBucket Cloud to use API 2.0 URLs.

chuckries:
BitBucket Cloud 2FA does not work in VS for a few reasons. The first is
that Basic HTTP Auth is completely disabled when 2FA is enabled on the
account. Falling back to OAuth authentication works, but only for the
official REST API. Currently, BitBucket Source Link produces 'web
browser' URL's, using the
https://bitbucket.org/{user}/{repo}/raw/{commit}/{path} format. This is
not compatible with OAuth authentication.

By changing the URL's produced to the 2.0 REST API format, this should
allow VS to succesfully authenticate to private BitBucket Cloud Repo's
when using 2FA.

### DIFF
--- a/src/SourceLink.Bitbucket.Git/GetSourceLinkUrl.cs
+++ b/src/SourceLink.Bitbucket.Git/GetSourceLinkUrl.cs
@@ -71,7 +71,7 @@ namespace Microsoft.SourceLink.Bitbucket.Git
 
         private static string BuildSourceLinkUrlForCloudEdition(Uri contentUri, string relativeUrl, string revisionId)
         {
-            // change bitbuket.org to api.bitbucket.org
+            // change bitbucket.org to api.bitbucket.org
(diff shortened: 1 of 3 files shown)


tmat:
@adam-sajdak FYI

src/SourceLink.Bitbucket.Git/GetSourceLinkUrl.cs
```diff
@@ -71,8 +71,13 @@ private Version GetBitbucketEnterpriseVersion(ITaskItem hostItem)
 
         private static string BuildSourceLinkUrlForCloudEdition(Uri contentUri, string relativeUrl, string revisionId)
         {
-            return UriUtilities.Combine(UriUtilities.Combine(contentUri.ToString(), relativeUrl),
-                "raw/" + revisionId + "/*");
+            // change bitbuket.org to api.bitbucket.org
```
tmat: Typo: `bitbuket.org`

PR #463, opened 2019-10-25
Normalize git commondir path.

AustinWise:
The constructor for GitRepositoryLocation asserts that this path is
normalized. On my machine, a git worktree created by Git 2.22.0 does not
result in path that satisfies this assert.

Fixes https://github.com/dotnet/sourcelink/issues/473

### DIFF
--- a/src/Microsoft.Build.Tasks.Git/GitDataReader/GitRepository.cs
+++ b/src/Microsoft.Build.Tasks.Git/GitDataReader/GitRepository.cs
@@ -491,6 +491,8 @@ namespace Microsoft.Build.Tasks.Git
                 try
                 {
                     commonDirectory = Path.Combine(directory, File.ReadAllText(commonLinkPath).TrimEnd(CharUtils.AsciiWhitespace));
+                    // Normalize relative paths. For example, git worktrees typically have "../.." in this file.
+                    commonDirectory = Path.GetFullPath(commonDirectory);
                 }
                 catch
                 {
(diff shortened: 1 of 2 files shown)


AustinWise:
Specifically, the commondir file contains `../..`. In my limited testing this only effects debug builds. Release builds without `Debug.Assert` seem happy with the relative path.

src/Microsoft.Build.Tasks.Git/GitDataReader/GitRepository.cs
```diff
@@ -491,6 +491,8 @@ private static bool IsGitDirectory(string directory, out string commonDirectory)
                 try
                 {
                     commonDirectory = Path.Combine(directory, File.ReadAllText(commonLinkPath).TrimEnd(CharUtils.AsciiWhitespace));
+                    // noralize relative paths
```
tmat: Typo in comment. I don't think the comment is needed since it's obvious what `GetFullPath` does. Also, you can simplify by calling `Path.GetFullPath()` directly on the previous line `Path.GetFullPath(Path.Combine(...))`

AustinWise:
Ok, I will update with a test and a fix for the typo.

AustinWise:
@tmat I added tests and fixed the comment. I was not sure if you want me to replace the existing test or add a new one, so for now I added a new one. The existing test creates a git layout that does not match the [repository specification](https://git-scm.com/docs/gitrepository-layout#Documentation/gitrepository-layout.txt-worktrees) nor the layout `git worktree` currently creates. So perhaps the old test should be replaced with the new test?

tmat:
Thanks!

PR #471, opened 2019-10-31
LOC CHECKIN | dotnet/sourcelink master | 20191031

v-zbsail:


tmeschter:
@tmat Does anything further need to be done with this? That is, does it need to be inserted into the VS repo for 16.4 or anything like that?

tmat:
Nope.

Issue #239, opened 2019-02-01
Question: Is SourceLink available using a Bonobo git server


robinabganpat:
I would like to know whether it's possible to use SourceLink in combination with a on premise Bonobo Git Server. I would like to be able to step-through my NuGet packages (which are using the .NET Core framework). 
As far as I know this is currenlty only possible using SourceLink. Would this require me to migrate my Git repos to a Git Server that's supported by SourceLink?

tmat:
Source Link is not limited to any specific source control server. The only requirement on the server is that it supports fetching raw source files given a repository url, commit sha an a relative path within the repository in some form. If the format of the raw URL matches GitHub Enterprise, GitLab or BitBucket you might be able to just reuse the implementation for these providers. If not you can implement a custom one. To do so you can start by copying one of the projects in this repository, e.g. 
https://github.com/dotnet/sourcelink/tree/master/src/SourceLink.Bitbucket.Git and then tweak the function that constructs the raw source file URL: https://github.com/dotnet/sourcelink/blob/master/src/SourceLink.Bitbucket.Git/GetSourceLinkUrl.cs#L20

robinabganpat:
> Source Link is not limited to any specific source control server. The only requirement on the server is that it supports fetching raw source files given a repository url, commit sha an a relative path within the repository in some form. If the format of the raw URL matches GitHub Enterprise, GitLab or BitBucket you might be able to just reuse the implementation for these providers. If not you can implement a custom one. To do so you can start by copying one of the projects in this repository, e.g.
> https://github.com/dotnet/sourcelink/tree/master/src/SourceLink.GitLab and then tweak the function that constructs the raw source file URL: https://github.com/dotnet/sourcelink/tree/master/src/SourceLink.GitLab/GetSourceLinkUrl.cs#L20

Thank you, that clarifies a lot. Only thing I'm wondering about is the way authentication is handled here. Accessing source files from my private repository requires logging in on the Git Server.

tmat:
As long as Bonobo supports basic HTTP auth the VS debugger will be able to authenticate.

robinabganpat:
Is there any documentation available of transforming the GitLab project into a custom project?
I currently have https://github.com/dotnet/sourcelink/tree/master/src/SourceLink.GitLab checked out and added it as a reference to my library project, which I want to pack as a NuGet package. Would this also be okay? Or do I first have to make the custom SourceLink.Bonobo available as a NuGet Package?
[...]

PR #505, opened 2019-11-25
Feature: GitWeb Source Link provider

Glen-Nicol-Garmin:
Hi there, I work for Garmin and we use an internal GitWeb server that doesn't work with any of the existing providers. I created one that should work for many other companies in a similar situation that don't use cloud providers.

I saw in #239 that the suggested solution for a provider that is not currently supported is to copy and edit the GitLab ones. So that is what I did.


tmat:
Cools. Thanks for the PR, much appreciated! We are currently working on releasing first stable version of SourceLink packages, so I'm gonna wait with merging this until after we finish that process.

Glen-Nicol-Garmin:
Sounds good. Is there an estimate on how long that will take?

tmat:
A couple of days :)

src/SourceLink.GitWeb/GetSourceLinkUrl.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            var trimLeadingSlash = relativeUrl.TrimStart('/', '\\');
+            var trimmedContentUrl = contentUri.ToString().TrimEnd('/', '\\');
+
+            /* p = project/path
+             * a = action
+             * hb = SHA/revision
+             * f = repo file path
+             */
+            var gitwebRawUrl = $"{trimmedContentUrl}/?p={trimLeadingSlash}.git;a=blob_plain;hb={revisionId};f=*";
```
tmat: >trimmedContentUrl [](start = 34, length = 17)

You can use `UriUtilities.Combine` helper to combine URL parts: `UriUtilities.Combine(contentUri.ToString(), $"?p=...")`

src/SourceLink.GitWeb/GetSourceLinkUrl.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// string "N/A".
+    /// </summary>
+    public sealed class GetSourceLinkUrl : GetSourceLinkUrlGitTask
+    {
+        protected override string HostsItemGroupName => "SourceLinkGitWebHost";
+        protected override string ProviderDisplayName => "GitWeb";
+
+        protected override string BuildSourceLinkUrl(Uri contentUri, Uri gitUri, string relativeUrl, string revisionId, ITaskItem hostItem)
+        {
+            var trimLeadingSlash = relativeUrl.TrimStart('/', '\\');
```
tmat: >'\\' [](start = 62, length = 4)

It shouldn't be necessary to trim backslash.

src/SourceLink.GitWeb/build/SourceLink.GitWeb.targets
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
```
tmat: The file name is missing `Microsoft.` prefix. That's probably why the build is failing.

src/SourceLink.GitWeb/buildMultiTargeting/SourceLink.GitWeb.targets
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
```
tmat: The file name is missing `Microsoft.` prefix. That's probably why the build is failing.

src/SourceLink.GitWeb/Microsoft.SourceLink.GitWeb.nuspec
```diff
@@ -0,0 +1,17 @@
...
+  <metadata>
+    $CommonMetadataElements$
+    <dependencies>
+      <dependency id="Microsoft.SourceLink.Common" version="$Version$" />
+      <dependency id="Microsoft.Build.Tasks.Git" version="$Version$" />
+    </dependencies>
+  </metadata>
+  <files>
+    $CommonFileElements$
+    <file src="**\SourceLink.GitWeb.*" target="tools" />
```
tmat: `**\Microsoft.SourceLink.GitWeb.*`

src/SourceLink.GitWeb/GetSourceLinkUrl.cs
```diff
@@ -0,0 +1,35 @@
...
+            var trimmedContentUrl = contentUri.ToString().TrimEnd('/', '\\');
+
+            /* p = project/path
```
tmat: >/* [](start = 12, length = 2)

Nit (style): use `//` comments.

src/SourceLink.GitWeb/GetSourceLinkUrl.cs
Glen-Nicol-Garmin: Done

src/SourceLink.GitWeb/GetSourceLinkUrl.cs
Glen-Nicol-Garmin: Done

src/SourceLink.GitWeb/build/SourceLink.GitWeb.targets
Glen-Nicol-Garmin: Whoops. Good catch. Guess I missed a couple when I decided to go back to the Microsoft namespace.
Done.

src/SourceLink.GitWeb/buildMultiTargeting/SourceLink.GitWeb.targets
Glen-Nicol-Garmin: Done

src/SourceLink.GitWeb/Microsoft.SourceLink.GitWeb.nuspec
Glen-Nicol-Garmin: Done

src/SourceLink.GitWeb/GetSourceLinkUrl.cs
Glen-Nicol-Garmin: Codemaid reformats // comments to shorten them which is why I used those. I have swapped them for you though.

Glen-Nicol-Garmin:
I'm looking into integration tests now. Will be a separate commit than the above review comments.

tmat:
@Glen-Nicol-Garmin Will you be able to write the integration test? I can help if needed.

Re PrivateRepositoryUrl - See https://github.com/dotnet/sourcelink/tree/master/docs#publishrepositoryurl

tmat:
Re SSH - the default implementation is provided by TranslateRepositoryUrlsGitTask, from which TranslateRepositoryUrls derives. If the default implementation works then TranslateRepositoryUrls can be empty. AzureDevOpsServer uses a different structure for SSH URL than it uses for HTTP URL, so it needs to override:

https://github.com/dotnet/sourcelink/blob/master/src/SourceLink.AzureDevOpsServer.Git/TranslateRepositoryUrls.cs#L14

Glen-Nicol-Garmin:
I can do the work for the tests but I am still a bit confused on how those properties should be set. Sounds like in this case they should both be left as SSH urls? I gather that because from my interpretation those properties are supposed to be the URL for the source control software to do a clone/checkout and gitweb only does SSH (as far as I can tell).

tmat:
> Sounds like in this case they should both be left as SSH urls? I gather that because from my interpretation those properties are supposed to be the URL for the source control software to do a clone/checkout and gitweb only does SSH (as far as I can tell).

They need to be translated to `https` URLs. `RepositoryUrl` is embedded in a `.nuspec` by NuGet pack task. See https://github.com/dotnet/sourcelink/blob/master/src/SourceLink.Git.IntegrationTests/GitHubTests.cs#L267

Glen-Nicol-Garmin:
When I read the nuspec docs I interpret that [property as the clone URL](https://docs.microsoft.com/en-us/nuget/reference/nuspec#repository).

> This should be a publicly available url that can be invoked directly by a version control software.

Which in our case is the SSH URL. We can't clone from a https URL. If SourceLink requires it to be a HTTPS url then fine, but we should add documentation for the divergence.

tmat:
> We can't clone from a https URL

Can you elaborate on why it is not possible to clone using https URL?

Glen-Nicol-Garmin:
Our git server only supports the SSH protocol that I am aware of.  When you go to a gitweb page it lists the SSH url to clone it.

tmat:
That seems odd. Have you tried the https version of the URI? Source Link needs HTTPs support for at least fetching individual raw source files. It would be surprising if you could do that but couldn't clone the repo using HTTPs.

Glen-Nicol-Garmin:
I thought the point of source link was to translate between repo Urls to content urls where the files can be found and downloaded? We have content over HTTPS.

And no simply changing the protocol of the URI does not work.

tmat:
> I thought the point of source link was to translate between repo Urls to content urls where the files can be found and downloaded? We have content over HTTPS.

That is correct. It's just surprising that you wouldn't have HTTPS URL that would work for clone. I have not seen that before -- GitHub, Azure DevOps, GitLab, BitBucket all allow both HTTPS and SSH (and also git protocol). I guess GitWeb can be configured to only support SSH? 

If that's the case we can either skip the translation completely by not setting `SourceControlManagerUrlTranslationTargets` in the GitWeb targets file and removing the translation task, or overriding some of the methods on the translation task class to not translate ssh to https.

tmat:
BTW, the reason we translate SSH and git protocols to HTTPS is to normalize the build results. You can imagine different build environments choosing different protocols for cloning the repo. The result of the build from all these environments will be the same, regardless of how they cloned the repo.

tmat:
How about merging this PR and leaving the integration tests and SSH translation for a follow up?

Glen-Nicol-Garmin:
I think I need to wrap my head around the SSH translation before I really want to consider it done and ready for consumption. I should be able to get to that next week.

tmat:
BTW, I'm going to merge https://github.com/dotnet/sourcelink/pull/546 which might cause your PR to fail build due to missing nullable annotations. Should be easy to add though.

Glen-Nicol-Garmin:
So funny story, I tried removing the translation step altogether (commented out the using Task, the property you mentioned, and the target) still got https URLS. And then I tried overriding all those functions in TranslateRepositoryUrlsGitTask. Neither works in the integration tests. I am thinking I may not have something setup right for those target files.

Glen-Nicol-Garmin:
Any leads or ideas on why changing those files isn't affecting the integration tests?

Glen-Nicol-Garmin:
I was able to figure it out. Did not know the integration tests rely on the nuget packages which are not built when you build the solution.

tmat:
Oh, sorry about that! Yes, the build doesn't run `pack` by default. You need to run `build -pack` from command line. I think it would make sense to build the packages by default. It would be easier to work with.

Glen-Nicol-Garmin:
No worries, would you like me to rebase onto master before you review the latest changes?

tmat:
@Glen-Nicol-Garmin Yes, please rebase.

Glen-Nicol-Garmin:
K, done. You should look at the changes I made to the common translation Task to facilitate checking for supported protocols.

src/Common/TranslateRepositoryUrlGitTask.cs
```diff
@@ -47,26 +50,45 @@ private void ExecuteImpl()
...
                             uri.Scheme.Equals("ssh", StringComparison.OrdinalIgnoreCase) ? TranslateSshUrl(uri) :
-                            uri.Scheme.Equals("git", StringComparison.OrdinalIgnoreCase) ? TranslateGitUrl(uri) : null) ?? url;
+                            uri.Scheme.Equals("git", StringComparison.OrdinalIgnoreCase) ? TranslateGitUrl(uri) : null) ?? uri.OriginalString;
                 }
 
-                return url;
+                return uri.OriginalString;
+            }
+
+            var providerName = this.GetType().Namespace;
```
tmat: > var providerName  [](start = 11, length = 18)

This is more complicated than it seems necessary. `isMatchingHostUri` will only return true for URLs specific to the provider (e.g. GitWeb). Thus the previous version of the code only translated URIs that belong to the specific provider. For GitWeb such URLs would only be SSH URLs, unless someone is tweaking the msbuild properties/items in an unsupported way. We want to handle that case gracefully as well, of course, and report a reasonable error. It seems that it would suffice to throw NotSupportedException from TranslateXxxUrl as you already do and wrap the code that calls to `translate` in `try-catch (NotSupportedException e) { Log.LogError(e.Message) }`. We don't need to check up-front for supported/unsupported schemes as the common case would not throw.

Glen-Nicol-Garmin: K, I thought about that but preferred to avoid exceptions if possible. But you're right that is vastly simpler. 

Before I do, should I use Resources for the exception string? And if so, how do I reference those common resources in tests? The ones I did add didn't show up on Common.Resources and I couldn't access CommonResources in tests.

tmat: Yes, use Resources for the exception string. You'll need to add InternalsVisibleTo here: https://github.com/dotnet/sourcelink/blob/master/src/SourceLink.Common/Microsoft.SourceLink.Common.csproj#L19

tmat: Actually, no need for that. Your resource string won't be in the Common library as you're constructing the exception with the message in GitWeb. In the common library you can just write out e.Message.

Glen-Nicol-Garmin: ... My goodness, duh! I'll go back to the C# basics course now. Thanks!

Glen-Nicol-Garmin: So should I just use a string or should I add resources to my provider?

tmat: Please add a resource.

Glen-Nicol-Garmin: How can I do that without overriding the common ones? When I add a Resources.resx file to the project the files in the xlf directory change and no longer have the resources that it inherited. Do I just need to copy the common ones into my project too?

tmat: Not sure I understand. `SourceLink.AzureRepos.Git` project adds its own Resources.resx and also uses the common ones like so:
https://github.com/dotnet/sourcelink/blob/master/src/SourceLink.AzureRepos.Git/Microsoft.SourceLink.AzureRepos.Git.csproj#L22

Glen-Nicol-Garmin: I guess what's confusing is that resources like 'ValueOfWithIdentityIsInvalid' and 'AtLeastOneRepositoryHostIsRequired' are defined in the CommonResources.resx and then each project has a xlf directory with translations of those.

But once I created a Resources.resx file in the gitweb project I lost all the CommonResources translations in those files. In the case of the AzureRepos project you mention its [Resources.resx](https://github.com/dotnet/sourcelink/blob/master/src/SourceLink.AzureRepos.Git/Resources.resx#L120) seems to duplicate some of the resources from Common. 

So once a provider has its own Resources does it need to duplicate the common ones it uses?

tmat: The duplication shouldn't be necessary, seems like it's there by mistake.

Glen-Nicol-Garmin: K, simplified logic, added a resource to the gitweb project and reset the changes to the CommonResources

Glen-Nicol-Garmin:
Also, as an aside maybe this belongs in a separate issue, whenever I run the tests an entry like this is added to my user PATH variable:
```
C:\Users\<username>\AppData\Local\Temp\SourceLinkTests\40f584a7-8ac8-427e-a078-05cd8fe22c08\.home\.dotnet\tools
```

These are never cleaned up. And it actually ended up breaking a few things on my system in a subtle way. 

For instance, my user TEMP directory wouldn't get set so all my programs would try to write to the system TEMP directory at C:\windows\TEMP. Which usually results in a Access denied error.

It also ended up causing my PATH to not be completely evaluated so things like code wouldn't run from the command line.

I guess with enough entries it overflows some max string limit in Windows and there is no error dialog or indicator anywhere when this happens.

tmat:
@Glen-Nicol-Garmin I have pushed a commit that removes Designer.cs file. We use a custom target to generate C# source from resx file that does not need a checked-in file.

src/SourceLink.GitWeb.UnitTests/GetSourceLinkUrlTests.cs
```diff
@@ -0,0 +1,58 @@
...
+
+            var task = new GetSourceLinkUrl()
+            {
+                BuildEngine = engine,
+                SourceRoot = new MockItem("/src/", KVP("RepositoryUrl", "ssh://git@src.intranet.company.com/root_dir_name/sub_dirs/reponame.git" + s1), KVP("SourceControl", "git"), KVP("RevisionId", "0123456789abcdefABCDEF000000000000000000")),
+                Hosts = new[]
+                {
+                    // NOTE: i don't know what the spec parameter is for. but for all the other
+                    // tests like this for various providers it is the domain of the Repo URL.
+                    new MockItem("src.intranet.company.com", KVP("ContentUrl", "https://src.intranet.company.com/gitweb" + s2)),
```
tmat: >new MockItem("src.intranet.company.com" [](start = 20, length = 39)

It is the host domain that needs to match the domain in the SourceRoot.

tmat:
@Glen-Nicol-Garmin Thanks for the contribution!

Glen-Nicol-Garmin:
Huzzah! That is great news! Before I forget, should I create an issue for that PATH bug in tests I mentioned?

tmat:
@Glen-Nicol-Garmin No need. This should take care of it: https://github.com/dotnet/sourcelink/pull/578.

tmat:
@Glen-Nicol-Garmin The package is now available in the [Azure feed](https://dev.azure.com/dnceng/public/_packaging?_a=package&feed=dotnet-tools&package=Microsoft.SourceLink.GitWeb&protocolType=NuGet&version=1.1.0-beta-20117-01):
Microsoft.SourceLink.GitWeb 1.1.0-beta-20117-01

Let me know if it works in a real-world project. Once you confirm I'll add a mention to Readme.md

Glen-Nicol-Garmin:
Just tested it out. It works!

src/Common/TranslateRepositoryUrlGitTask.cs
```diff
@@ -66,7 +67,16 @@ static bool isMatchingHostUri(Uri hostUri, Uri uri)
...
-            TranslatedRepositoryUrl = translate(RepositoryUrl);
+            try
+            {
+                TranslatedRepositoryUrl = translate(RepositoryUrl);
+            }
+            catch (NotSupportedException e)
+            {
+                Log.LogError(e.Message);
+                return;
+            }
```
dfev77, line 78: Is it ok to just log and return? I'm thinking that the error might be observer only a few week/months later when once debugs the code and sees that source linking doesn't work.
Why not let the exception propagate?

Glen-Nicol-Garmin: I am fairly hazy on the specifics now, but I think logging the error results in an MsBuild error that can't be suppressed. Is the concern that since it is a post build action that error may not be noticed since the files will be on disk? I think the exception being thrown would have the same limitation. 

I don't remember the specific use case where the NotSupportedException was thrown to say whether or not the message is helpful enough to tell the user that their URL scheme/format is not supported by Sourcelink. But I thought those were the errors being thrown by the individual translators when they didn't support the repo url.

Issue #576, opened 2020-02-16
Relative git submodules are not indexed


NN---:
Given git submodule with relative path:

.gitmodules
```ini
[submodule "a"]
   path = a
   url = ../a
```

SourceLink produces a warning and doesn't index files.

It doesn't produce a warning when a full URL is specified.
```ini
[submodule "a"]
   path = a
   url = htttp://github.com/company/a
```

tmat:
What warning is reported?

This scenario is covered by test [`GetSourceRoots_RelativeSubmodulePaths`](https://github.com/dotnet/sourcelink/blob/f61f34798be0641e4bbfa3a50174c8ee1f4f7ee5/src/Microsoft.Build.Tasks.Git.UnitTests/GitOperationsTests.cs#L504).

NN---:
```
packages\Microsoft.Build.Tasks.Git.1.0.0\build\Microsoft.Build.Tasks.Git.targets(24,5): warning : Repository 'c:\project\a.git' has no 
remote.
[c:\project\project.vcxproj]

packages\Microsoft.Build.Tasks.Git.1.0.0\build\Microsoft.Build.Tasks.Git.targets(24,5): warning : Repository 'c:\project\b' has no remote.
[c:\project\project.vcxproj]
```

tmat:
Can you share the structure of all the repositories involved? That is the content of their `.git` directories (just the files directly in those directories) and their locations on disk?
Alternatively, steps to reproduce.

tmat:
Does the repository in `c:\project` have a remote set up?

NN---:
Check this repository: https://github.com/NN---/sourcelinkmain

```
git clone https://github.com/NN---/sourcelinkmain sourcelinkmain
cd sourcelinkmain
git submodule update
nuget restore sourcelinkmain.sln
devenv sourcelinkmain.sln /build
```

```
1>C:\sourcelinkmain\packages\Microsoft.Build.Tasks.Git.1.0.0\build\Microsoft.Build.Tasks.Git.targets(24,5): warning : Repository 'C:\Projects\tests\sourcelinka.git' has no remote.
1>C:\sourcelinkmain\packages\Microsoft.Build.Tasks.Git.1.0.0\build\Microsoft.Build.Tasks.Git.targets(24,5): warning : Repository 'C:\Projects\tests\sourcelinkb' has no remote.
```

Zastai:
I have the same issue.

Given
[...]

PR #583, opened 2020-02-22
GH-576: Fix relative submodule URL handling.

Zastai:
They should be resolved against the main repository's URL, not the working directory.

### DIFF
--- a/src/Microsoft.Build.Tasks.Git/GitOperations.cs
+++ b/src/Microsoft.Build.Tasks.Git/GitOperations.cs
@@ -158,7 +158,7 @@ namespace Microsoft.Build.Tasks.Git
         internal static Uri? NormalizeUrl(GitRepository repository, string url)
         {
             // Git (v2.23.0) treats local relative URLs as relative to the working directory.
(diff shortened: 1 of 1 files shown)


Zastai:
An alternative might be to ignore the URL in .gitmodules, and instead use the branch specification to get the repository url from the module's own config file, if it exists.

Zastai:
Currently I essentially inlined a NormalizeUrl overload, tweaking the root path argument.
It may be more desirable to create another overload (or a NormalizeSubmoduleUrl() method) instead.

Zastai:
(Update: restored accidentally-removed whitespace, and corrected the issue name in a comment.)

tmat:
Thanks for the proposed change. I see what the problem is now. I think however that we need to check `submodule.<name>.url` configuration variable to find the URL for the submodule (see https://git-scm.com/docs/gitsubmodules). This variable is calculated based on the entry in `.gitmodules` by `git submodule init` and will be present for initialized submodules. We can skip uninitialized submodules as they won't have source code contributing to the build anyways. I'm gonna send a PR implementing this approach.

tmat:
Superseded by https://github.com/dotnet/sourcelink/pull/587.

PR #600, opened 2020-04-04
List GitWeb in README

tmat:


tmat:
@Glen-Nicol-Garmin FYI

Glen-Nicol-Garmin:
I suppose we should probably mention these just in case the defaults don't work.
```xml
<ItemGroup>
  <SourceLinkGitWebHost Include='your.gitweb.clone.domain.com' ContentUrl='https://gitweb.website.domain.com/gitweb'/>
</ItemGroup>
```

I think in 99% of cases the repo's SSH domain will be the same as the website content domain though. So Maybe not?

Thanks for doing that, I forgot about the readme!

PR #605, opened 2020-04-19
Add support for C++ libs into sourcelink

nickdalt:
Modify Microsoft.SourceLink.Common as follows:-

- Targets files generates sourcelink file only for C++ static library projects
- Targets file has new target to merge the sourcelink files associated with each lib into the master sourcelink file. Only executed for dll/exe C++ projects
- New task to perform the merge of sourcelink files

Makes the assumption that the for a C++ static lib the sourcelink file will be in the same directory with the extension sourcelink.json

Possible additional tasks for maintainer:-
- I am not that familiar with MSBuild so I may be using the wrong properties, conditions etc
- Localisation of log messages for task

See also issue#580

### DIFF
--- a/src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
+++ b/src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
@@ -3,6 +3,7 @@
   <Import Project="InitializeSourceControlInformation.targets"/>
   
   <UsingTask TaskName="Microsoft.SourceLink.Common.GenerateSourceLinkFile" AssemblyFile="$(_MicrosoftSourceLinkCommonAssemblyFile)"/>
+  <UsingTask TaskName="Microsoft.SourceLink.Common.MergeSourceLinkFiles" AssemblyFile="$(_MicrosoftSourceLinkCommonAssemblyFile)"/>
 
   <Target Name="_SetSourceLinkFilePath">
     <PropertyGroup>
@@ -28,9 +29,17 @@
           Condition="'$(SourceRootMappedPathsFeatureSupported)' == 'true'"/>
 
   <!-- 
+    Add compiler targets: C++ generates sourcelink file only for static libs.
+  -->
+  <PropertyGroup Condition="'$(Language)' == 'C++' and $(ConfigurationType) == 'StaticLibrary'">
+    <_GenerateSourceLinkFileBeforeTargets>BeforeClCompile</_GenerateSourceLinkFileBeforeTargets>
+    <_GenerateSourceLinkFileDependsOnTargets/>
+  </PropertyGroup>
+
+  <!-- 
     Add compiler targets: C++ generates PDB with SourceLink in Link phase.
   -->
-  <PropertyGroup Condition="'$(Language)' == 'C++'">
+  <PropertyGroup Condition="'$(Language)' == 'C++' and $(ConfigurationType) != 'StaticLibrary'">
     <_GenerateSourceLinkFileBeforeTargets>Link</_GenerateSourceLinkFileBeforeTargets>
     <_GenerateSourceLinkFileDependsOnTargets>ComputeLinkSwitches</_GenerateSourceLinkFileDependsOnTargets>
   </PropertyGroup>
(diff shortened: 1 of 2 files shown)


tmat:
Thanks for the PR. 

I'm not sure there is much benefit in merging the Source Link files. VC++ [linker](https://docs.microsoft.com/en-us/cpp/build/reference/sourcelink?view=vs-2019) supports multiple `/sourcelink` entries: `/link /sourcelink <sourcelinkfile1.json> /sourcelink <sourcelinkfile2.json>`. If multiple are specified the linker creates named streams called `sourcelink$1`, `sourcelink$2`, etc. and the debugger recognizes these.

This was specifically designed to remove complexity of merging JSON from build systems such as CMake.

nickdalt:
OK, I hadn't realised that /sourcelink could be used multiple times. I have updated the code to used /sourcelink with each discovered .sourcelink.json file.

src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
```diff
@@ -27,10 +28,18 @@
           DependsOnTargets="InitializeSourceRootMappedPaths"
           Condition="'$(SourceRootMappedPathsFeatureSupported)' == 'true'"/>
 
+  <!-- 
+    Add compiler targets: C++ generates sourcelink file only for static libs.
+  -->
+  <PropertyGroup Condition="'$(Language)' == 'C++' and $(ConfigurationType) == 'StaticLibrary'">
```
tmat, line 34: >$(ConfigurationType) [](http://example.com/codeflow?start=55&length=20)

Missing `'` around `$(ConfigurationType)`.

src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
```diff
@@ -27,10 +28,18 @@
...
+    <_GenerateSourceLinkFileBeforeTargets>BeforeClCompile</_GenerateSourceLinkFileBeforeTargets>
+    <_GenerateSourceLinkFileDependsOnTargets/>
+  </PropertyGroup>
+
   <!-- 
     Add compiler targets: C++ generates PDB with SourceLink in Link phase.
   -->
-  <PropertyGroup Condition="'$(Language)' == 'C++'">
+  <PropertyGroup Condition="'$(Language)' == 'C++' and $(ConfigurationType) != 'StaticLibrary'">
```
tmat, line 42: >$(ConfigurationType) [](http://example.com/codeflow?start=55&length=20)

Missing `'` around `$(ConfigurationType)`.

src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
```diff
@@ -45,20 +54,27 @@
...
-    <ItemGroup Condition="'$(Language)' == 'C++'">
+    <!-- Locate any additional sourcelink files associated with static libs -->
+    <Microsoft.SourceLink.Common.FindAdditionalSourceLinkFiles SourceLinkFile="$(SourceLink)" AdditionalLibraryDirectories="%(Link.AdditionalLibraryDirectories)" AdditionalDependencies="%(Link.AdditionalDependencies)" ImportLibraries="@(ProjectReferenceToLink)">
+      <Output TaskParameter="AllSourceLinkFiles" ItemName="SourceLinks" />
+    </Microsoft.SourceLink.Common.FindAdditionalSourceLinkFiles>
+
+    <!-- C++ Link task currently doesn't recognize SourceLink property only add this for non-static libs since lib doesn't
+         understand /sourcelink 
+    -->
+    <ItemGroup Condition="'$(Language)' == 'C++' and $(ConfigurationType) != 'StaticLibrary'">
```
tmat, line 75: >$(ConfigurationType) [](http://example.com/codeflow?start=53&length=20)

Missing `'` `'` around `$(ConfigurationType)`.

src/SourceLink.Common/FindAdditionalSourceLinkFiles.cs
```diff
@@ -0,0 +1,106 @@
...
+        public string[]? AllSourceLinkFiles { get; set; }
+
+        public override bool Execute()
+        {
+            List<string> allSourceLinkFiles = new List<string>();
+            allSourceLinkFiles.Add(SourceLinkFile);
+
+            try
+            {
+                //// Throughout we expect that the sourcelink files for a lib is alongside
```
tmat, line 48: >s [](http://example.com/codeflow?start=66&length=1)

typo: `file`

src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
```diff
@@ -45,20 +54,27 @@
...
       <Link Update="@(Link)">
-        <AdditionalOptions>%(Link.AdditionalOptions) /sourcelink:"$(SourceLink)"</AdditionalOptions>
+        <AdditionalOptions>%(Link.AdditionalOptions) @(SourceLinks->'/sourcelink:&quot;%(Identity)&quot;', ' ')</AdditionalOptions>
```
tmat, line 77: >Identity [](http://example.com/codeflow?start=89&length=8)

The values need to be escaped before they can be used as command line arguments. Perhaps, for simplicity of msbuild, FindAdditionalSourceLinkFiles could instead produce the command line string instead of individual items?

tmat:
@nickdalt Thanks for the PR - sorry for the delay in review. Was busy with other things. 

Any ideas how we can test this reasonably?

tmat (changes requested):
Command line switches need escaping and we also need a test.

src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
nickdalt: Done

src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
nickdalt: Done

src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
nickdalt: Done

src/SourceLink.Common/FindAdditionalSourceLinkFiles.cs
nickdalt: Fixed

nickdalt:
> Command line switches need escaping and we also need a test.

I am not clear what you want here

> Any ideas how we can test this reasonably?
> 
> I think we should at least have unit tests for FindAdditionalSourceLinkFiles task. End-to-end test might be too complicated since it would depend on VC++ compiler targets.

I can't see any easy way of testing end to end since that would require publishing the NuGet, consuming it in multiple C++ projects, building them, and finally checking that the pdb contains the sourcelink details for the libs.

I have however added unit tests for FindAdditionalSourceLinkFiles

sylveon:
I am interested in this - however my setup is a few static libraries packaged in a NuGet package and then acquired in another solution. I imagine in this setup, it would be required for the NuGet package to ship the relevant .sourcelink.json files, and then for the consumer of my package to also have the SourceLink NuGet package installed so that it can then be passed to `link.exe` during the final link to a .DLL file, correct?

Also, it seems SourceLink is disabled by default for IDE builds, but in a setup like the one mentioned here, we would need to at least get the SourceLink metadata from the static libraries into the final PDB even from IDE builds. Would we need to enable SourceLink as a whole for IDE builds or is this something this PR considers?

I guess, alternatively, we could tweak our NuGet package's `props` file to pass the relevant `/sourcelink` flags for the .lib files it ships, but would the linker be okay with potentially passing the same `sourcelink.json` file twice? For example if the consuming project also uses SourceLink.

tmat:
@nickdalt Sorry for the long delay. Switched focus to some other work. If you're still interested in bringing this PR forward, please rebase to the latest main.

PR #651, opened 2020-09-24
Add arcade-powered source-build basic configuration and patches

dagood:
This PR adds arcade-powered source-build configuration, so `./build.sh -c Release /p:ArcadeBuildFromSource=true` creates a source-build intermediate nuget package. It also adds patches that we have been maintaining in dotnet/source-build.

### DIFF
--- a/eng/SourceBuild.props
+++ b/eng/SourceBuild.props
@@ -1,0 +1,15 @@
+<Project>
+
+  <PropertyGroup>
+    <GitHubRepositoryName>sourcelink</GitHubRepositoryName>
+    <SourceBuildManagedOnly>true</SourceBuildManagedOnly>
+  </PropertyGroup>
+
+  <Target Name="GetRepoSourceBuildCommandConfiguration"
+          BeforeTargets="GetSourceBuildCommandConfiguration">
+    <PropertyGroup>
+      <InnerBuildArgs>$(InnerBuildArgs) /p:Pack=true</InnerBuildArgs>
+    </PropertyGroup>
+  </Target>
+
+</Project>
--- a/eng/SourceBuildPrebuiltBaseline.xml
+++ b/eng/SourceBuildPrebuiltBaseline.xml
@@ -1,0 +1,5 @@
+<UsageData>
+  <IgnorePatterns>
+    <UsagePattern IdentityGlob="*/*" />
+  </IgnorePatterns>
+</UsageData>
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -7,5 +7,10 @@
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>61cde6e8fb9d5c9790867b279deb41783a780cd8</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.SourceBuild.Intermediate.source-build-reference-packages" Version="5.0.0-alpha.1.20473.1">
+      <Uri>https://github.com/dotnet/source-build-reference-packages</Uri>
+      <Sha>def2e2c6dc5064319250e2868a041a3dc07f9579</Sha>
+      <SourceBuild RepoName="source-build-reference-packages" ManagedOnly="true" />
+    </Dependency>
   </ToolsetDependencies>
 </Dependencies>


src/Microsoft.Build.Tasks.Git/GitDataReader/GitRepository.cs
```diff
@@ -209,6 +209,11 @@ public static GitRepository OpenRepository(GitRepositoryLocation location, GitEn
         /// <returns>Null if the HEAD tip reference can't be resolved.</returns>
         internal string? ReadSubmoduleHeadCommitSha(string submoduleWorkingDirectoryFullPath)
         {
+            if (Directory.Exists(Path.Combine(submoduleWorkingDirectoryFullPath, ".git")))
+            {
+                return File.ReadAllText(Path.Combine(submoduleWorkingDirectoryFullPath, ".git", "HEAD"));
+            }
```
dagood, line 215: @crummel can you give some context on this patch? My understanding has been that a submodule by definition has a `.git` file that points to the superproject's Git data for that submodule. Maybe something's being treated as a submodule that shouldn't be?

dagood: This patch seems to be breaking tests around this scenario, which makes sense: https://github.com/dotnet/sourcelink/blob/8a3edd1902dbfe3adba65f22e3bb7aa2cc73e97f/src/Microsoft.Build.Tasks.Git.UnitTests/GitRepositoryTests.cs#L267-L269

I'm going to revert it for now.

dagood:
Looking at a few of the test failures, it looks like we're hitting the same problems as this plain old Arcade update PR: https://github.com/dotnet/sourcelink/pull/650.

@tmat can you take a look at https://github.com/dotnet/sourcelink/pull/650 to get sourcelink closer to what we need for source-build? Then we can get a cleaner read on PR validation in this PR.

eng/common/tools.ps1
```diff
@@ -312,9 +312,15 @@ function InitializeVisualStudioMSBuild([bool]$install, [object]$vsRequirements =
     return $global:_MSBuildExe
   }
 
-  $vsMinVersionReqdStr = '16.5'
+  # Minimum VS version to require.
```
tmat, line 315: ># Minimum VS version to require. [](start = 2, length = 32)

Shouldn't this be auto-updated by Maestro?

tmat: Seems like [https://github.com/dotnet/sourcelink/pull/650](https://github.com/dotnet/sourcelink/pull/650) is blocked currently. I'd prefer we resolve that issue and let Maestro update these files.

---
In reply to: [494513458](https://github.com/dotnet/sourcelink/pull/651#discussion_r494513458) [](ancestors = 494513458)

src/SourceLink.AzureDevOpsServer.Git/build/Microsoft.SourceLink.AzureDevOpsServer.Git.targets
```diff
@@ -63,7 +63,7 @@
     </Microsoft.SourceLink.AzureDevOpsServer.Git.TranslateRepositoryUrls>
 
     <ItemGroup>
-      <SourceRoot Remove="@(SourceRoot)"/>
+      <SourceRoot Remove="@(SourceRoot)" Condition="'@(_TranslatedSourceRoot)' != ''"/>
```
tmat, line 66: >" Condition="'@(_TranslatedSourceRoot)' != ''"/> [](start = 39, length = 48)

This is a product change. Is there an issue filed that this is fixing? We should also have a test covering this if this is indeed a product bug.

tmat (changes requested):
:clock1:

eng/common/tools.ps1
dagood: It was auto-updated by Darc in this case, in my first commit. And yeah, I pointed at https://github.com/dotnet/sourcelink/pull/650 in my comment on this PR. 😄

src/SourceLink.AzureDevOpsServer.Git/build/Microsoft.SourceLink.AzureDevOpsServer.Git.targets
dagood: @crummel do you have any extra info on this patch?

src/Microsoft.Build.Tasks.Git/GitDataReader/GitRepository.cs
crummel: The idea here was to use the [Minimal git repository metadata](https://github.com/dotnet/sourcelink/blob/master/docs/README.md) support.  Source-build doesn't have real repos (we purposely use non-git source directories to reduce confusion and tarball size), so we write out .git/HEAD to these folders to make sourcelink work.  This didn't work in submodules because sourcelink always treats submodules as having .gitdir redirects instead of real .git folders (which in a normal situation is true).  I'll take a look and try to figure out a clean fix.

src/SourceLink.AzureDevOpsServer.Git/build/Microsoft.SourceLink.AzureDevOpsServer.Git.targets
crummel: I remember that `TranslateRepositoryUrls` was returning an empty list for `TranslatedSourceRoot`s.  Then with the removal of the original SourceRoots, we ended up with no SourceRoots at all, which caused the usual "no SourceRoots" error.  I'll take a look at why this was happening.

dagood: Removed for now.

eng/common/tools.ps1
dagood: Rebased on top of the merge, now this PR just has the one commit adding basic arcade-powered-source build configuration.

dagood:
@tmat can you take another look?

I removed the patches, so this is now only the minimal infra. We can investigate further on the need for the patches and integrate changes as needed later after discussion.

src/SourceLink.AzureDevOpsServer.Git/build/Microsoft.SourceLink.AzureDevOpsServer.Git.targets
crummel: My test build looks good without this, so I think it was a temporary issue.

src/Microsoft.Build.Tasks.Git/GitDataReader/GitRepository.cs
crummel: Opened https://github.com/dotnet/sourcelink/pull/653 for a better version of this patch.

tmat (approved):
:shipit:

PR #653, opened 2020-09-30
Allow using .git directory instead of gitdir redirect in submodules.

crummel:
This is not a scenario for normal git repos, but is still legal, and the Arcade uberclone purposely uses this scenario for minimal SourceLink metadata.

### DIFF
--- a/src/Microsoft.Build.Tasks.Git.UnitTests/GitRepositoryTests.cs
+++ b/src/Microsoft.Build.Tasks.Git.UnitTests/GitRepositoryTests.cs
@@ -295,6 +295,7 @@ namespace Microsoft.Build.Tasks.Git.UnitTests
             {
                 "S10: 'sub10' 'http://github.com'",
                 "S11: 'sub11' 'http://github.com'",
+                "S6: 'sub6' 'http://github.com'",
                 "S9: 'sub9' 'http://github.com'"
             }, submodules.Select(s => $"{s.Name}: '{s.WorkingDirectoryRelativePath}' '{s.Url}'"));
 
@@ -311,8 +312,6 @@ namespace Microsoft.Build.Tasks.Git.UnitTests
               TestUtilities.GetExceptionMessage(() => File.ReadAllText(Path.Combine(workingDir.Path, "sub4", ".git"))),
               // Could not find a part of the path 'sub5\.git'.
               TestUtilities.GetExceptionMessage(() => File.ReadAllText(Path.Combine(workingDir.Path, "sub5", ".git"))),
-              // Access to the path 'sub6\.git' is denied
-              TestUtilities.GetExceptionMessage(() => File.ReadAllText(Path.Combine(workingDir.Path, "sub6", ".git"))),
               // The format of the file 'sub7\.git' is invalid.
               string.Format(Resources.FormatOfFileIsInvalid, Path.Combine(workingDir.Path, "sub7", ".git")),
               // Path specified in file 'sub8\.git' is invalid.
--- a/src/Microsoft.Build.Tasks.Git.UnitTests/GitRepositoryTests.cs
+++ b/src/Microsoft.Build.Tasks.Git.UnitTests/GitRepositoryTests.cs
@@ -344,14 +344,26 @@ namespace Microsoft.Build.Tasks.Git.UnitTests
             var gitDir = temp.CreateDirectory();
             var workingDir = temp.CreateDirectory();
 
-            var basicSubmoduleGitDir = temp.CreateDirectory();
+            var submoduleGitDir = temp.CreateDirectory();
 
-            var basicSubmoduleWorkingDir = workingDir.CreateDirectory("sub").CreateDirectory("abc");
-            basicSubmoduleWorkingDir.CreateFile(".git").WriteAllText("gitdir: " + basicSubmoduleGitDir.Path + "\t \v\f\r\n\n\r");
+            var submoduleWorkingDir = workingDir.CreateDirectory("sub").CreateDirectory("abc");
+            submoduleWorkingDir.CreateFile(".git").WriteAllText("gitdir: " + submoduleGitDir.Path + "\t \v\f\r\n\n\r");
 
-            var basicSubmoduleRefsHeadsDir = basicSubmoduleGitDir.CreateDirectory("refs").CreateDirectory("heads");
-            basicSubmoduleRefsHeadsDir.CreateFile("master").WriteAllText("0000000000000000000000000000000000000000");
-            basicSubmoduleGitDir.CreateFile("HEAD").WriteAllText("ref: refs/heads/master");
+            var submoduleRefsHeadsDir = submoduleGitDir.CreateDirectory("refs").CreateDirectory("heads");
+            submoduleRefsHeadsDir.CreateFile("master").WriteAllText("0000000000000000000000000000000000000000");
+            submoduleGitDir.CreateFile("HEAD").WriteAllText("ref: refs/heads/master");
+
+            var repository = new GitRepository(GitEnvironment.Empty, GitConfig.Empty, gitDir.Path, gitDir.Path, workingDir.Path);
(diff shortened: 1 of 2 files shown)


src/Microsoft.Build.Tasks.Git/GitDataReader/GitRepository.cs
```diff
@@ -209,6 +209,15 @@ public static GitRepository OpenRepository(GitRepositoryLocation location, GitEn
         /// <returns>Null if the HEAD tip reference can't be resolved.</returns>
         internal string? ReadSubmoduleHeadCommitSha(string submoduleWorkingDirectoryFullPath)
         {
+            // submodules don't usually have their own .git directories but can when Arcade uberclone
+            // was used.  Handle this case first since the other case throws.
```
dagood, line 213: Ah, I was skeptical since I'd never seen it, but you're right that submodules may have a `.git` directory rather than `.git` file:

https://git-scm.com/docs/gitsubmodules#_forms

> "Old-form" submodule: A working directory with an embedded .git directory, and the tracking gitlink and .gitmodules entry in the superproject. This is typically found in repositories generated using older versions of Git.

https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt-absorbgitdirs

> A repository that was cloned independently and later added as a submodule or old setups have the submodules git directory inside the submodule instead of embedded into the superprojects git directory.

I don't think uberclone actually does make "real" submodules registered with Git--it just clones a repo within a repo--but I don't think this distinction matters. The reason we're in this code in the first place is that sourcelink is assuming a repo-within-a-repo is always a submodule, it seems.

src/Microsoft.Build.Tasks.Git/GitDataReader/GitRepository.cs
```diff
@@ -209,6 +209,15 @@ public static GitRepository OpenRepository(GitRepositoryLocation location, GitEn
         /// <returns>Null if the HEAD tip reference can't be resolved.</returns>
         internal string? ReadSubmoduleHeadCommitSha(string submoduleWorkingDirectoryFullPath)
         {
+            // submodules don't usually have their own .git directories but can when Arcade uberclone
```
tmat, line 212: Remove any references to Arcade. This is not infrastructure code, this is a shipping product.

tmat (changes requested):
We need a test that validates this behavior.

It'd also be useful to have a sequence of git commands that create this kind of structure.

src/Microsoft.Build.Tasks.Git/GitDataReader/GitRepository.cs
crummel: Removed.

crummel:
I think I've addressed your comments @tmat, could you take a look when you have a chance?

crummel:
@tmat I think @MichaelSimons is running into this in another case, could you take a look at the PR again?

MichaelSimons:
@tmat, @crummel - Friendly ping.  Can we get a fix in?  This is needed to unblock the offline source-build builds for 6.0

src/Microsoft.Build.Tasks.Git/GitDataReader/GitRepository.cs
```diff
@@ -209,6 +209,18 @@ public static GitRepository OpenRepository(GitRepositoryLocation location, GitEn
         /// <returns>Null if the HEAD tip reference can't be resolved.</returns>
         internal string? ReadSubmoduleHeadCommitSha(string submoduleWorkingDirectoryFullPath)
         {
+            // Submodules don't usually have their own .git directories but this is still legal.
+            // This can occur with older versions of Git or other tools, or when a user clones one
+            // repo into another's source tree (but it was not yet registered as a submodule).
+            // See https://git-scm.com/docs/gitsubmodules#_forms for more details.
+            // Handle this case first since the other case throws.
+            var dotGitPath = Path.Combine(submoduleWorkingDirectoryFullPath, GitDirName);
```
tmat, line 217: >var dotGitPath = Path.Combine(submoduleWorkingDirectoryFullPath, GitDirName); [](http://example.com/codeflow?start=12&length=77)

Let's avoid some code duplication here:

```C#
var dotGitPath = Path.Combine(submoduleWorkingDirectoryFullPath, GitDirName);

var gitDirectory = 
    Directory.Exists(dotGitPath) ? dotGitPath : 
    File.Exists(dotGitPath) ? ReadDotGitFile(dotGitPath) :
    null;

if (gitDirectory == null || !IsGitDirectory(gitDirectory, out var commonDirectory))
{
    return null;
}

var resolver = new GitReferenceResolver(gitDirectory, commonDirectory);
return resolver.ResolveHeadReference();
```

src/Microsoft.Build.Tasks.Git.UnitTests/GitRepositoryTests.cs
```diff
@@ -344,17 +344,26 @@ public void GetSubmoduleHeadCommitSha()
...
+            basicSubmoduleRefsHeadsDir.CreateFile("master").WriteAllText("0000000000000000000000000000000000000000");
+            basicSubmoduleGitDir.CreateFile("HEAD").WriteAllText("ref: refs/heads/master");
+
+            // this is a unusual but legal case which can occur with older versions of Git or other tools.
+            // see https://git-scm.com/docs/gitsubmodules#_forms for more details.
+            var oldStyleSubmoduleWorkingDir = workingDir.CreateDirectory("old-style-submodule");
+            var oldStyleSubmoduleGitDir = oldStyleSubmoduleWorkingDir.CreateDirectory(".git");
+            var oldStyleSubmoduleRefsHeadDir = oldStyleSubmoduleGitDir.CreateDirectory("refs").CreateDirectory("heads");
+            oldStyleSubmoduleRefsHeadDir.CreateFile("branch1").WriteAllText("1111111111111111111111111111111111111111");
+            oldStyleSubmoduleGitDir.CreateFile("HEAD").WriteAllText("ref: refs/heads/branch1");
```
tmat, line 362: I'd prefer this case to be a separate test to keep the tests simpler.

src/Microsoft.Build.Tasks.Git/GitDataReader/GitRepository.cs
crummel: Changed as suggested.

src/Microsoft.Build.Tasks.Git.UnitTests/GitRepositoryTests.cs
crummel: Split this out into a separate test.

tmat (approved):
:shipit:

src/Microsoft.Build.Tasks.Git/GitDataReader/GitRepository.cs
```diff
@@ -209,8 +209,19 @@ public static GitRepository OpenRepository(GitRepositoryLocation location, GitEn
...
+
+            var gitDirectory =
+                Directory.Exists(dotGitPath) ? dotGitPath :
+                File.Exists(dotGitPath) ? ReadDotGitFile(dotGitPath) :
```
tmat, line 221: A test is failing - looking at it I think we should keep throwing if the file can't be read, rather then returning null. 

```C#
var gitDirectory = Directory.Exists(dotGitPath) ? dotGitPath : ReadDotGitFile(dotGitPath);

if (!IsGitDirectory(gitDirectory, out var commonDirectory))
```

tmat: With this change you'd still need to update the `Submodules_Errors` test - remove case S6:
```
[submodule ""S6""]             # sub6/.git is a directory, but should be a file
  path = sub6
  url = http://github.com
````

tmat: Or rather update it - since in this case the directory is not a valid git directory (it's empty) so failure is expected of some sort

PR #668, opened 2020-11-06
Fix broken link in README

0xced:
The link was pointing to the master branch so the line numbers have changed on master since then. The new link points to a tag to ensure that line numbers won't change in the future.


tmat:
Thanks!

PR #691, opened 2021-01-13
Try source build pool fix

dagood:
Try a fix for https://github.com/dotnet/source-build/issues/1795. I wasn't able to configure an AzDO pipeline to simulate the bad default situation--requires the old UI which didn't present the right pool options. But, a draft PR here should be fine to try it.


dagood:
Fix worked, job failed while creating intermediate nupkg due to license file validation.

dagood:
`--ci` causes `_RepositoryValidation` to run, which detects that the intermediate nupkg is generated with the MIT license, but this repo uses the Apache 2.0 license.

Can repro locally, will fix.

PR #699, opened 2021-02-01
Allow retrieval of the "relative file path" from the source link map

chuckries:
It can be helpful for consumers of the Source Link to know what portion
of the file path provided to the query is substituted into the URL. One
example is for building a heuristic file path on disk to store the
downloaded source link file to. This relative path can give end users
more context on where the file is located in the original repository.

This change add a new TryGetUri overload that provides the "relative
file path" as another out param. It also updates the tests to add
coverage for this.

### DIFF
--- a/src/SourceLink.Tools/SourceLinkMap.cs
+++ b/src/SourceLink.Tools/SourceLinkMap.cs
@@ -190,6 +190,24 @@ namespace Microsoft.SourceLink.Tools
 #endif
             out string? uri)
         {
+            return TryGetUri(path, out uri, out _);
+        }
+
+        /// <summary>
+        /// Maps specified <paramref name="path"/> to the corresponding URL.
+        /// </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="path"/> is null.</exception>
+        public bool TryGetUri(
+            string path,
+#if NETCOREAPP
+            [NotNullWhen(true)]
+#endif
+            out string? uri,
+#if NETCOREAPP
+            [NotNullWhen(true)]
+#endif
+            out string? relativeFilePath)
+        {
             if (path == null)
             {
                 throw new ArgumentNullException(nameof(path));
@@ -198,6 +216,7 @@ namespace Microsoft.SourceLink.Tools
             if (path.IndexOf('*') >= 0)
             {
                 uri = null;
+                relativeFilePath = null;
                 return false;
             }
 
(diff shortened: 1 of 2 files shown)


chuckries:
cc @tmat

src/SourceLink.Tools/SourceLinkMap.cs
```diff
@@ -209,20 +228,24 @@ private static bool TryParseEntry(string key, string value, out Entry entry)
...
+                        var escapedPath = string.Join("/", relativePath.Split('/').Select(Uri.EscapeDataString));
                         uri = mappedUri.Prefix + escapedPath + mappedUri.Suffix;
+                        relativeFilePath = relativePath;
                         return true;
                     }
                 }
                 else if (string.Equals(path, file.Path, StringComparison.OrdinalIgnoreCase))
                 {
                     Debug.Assert(mappedUri.Suffix.Length == 0);
+                    relativeFilePath = Path.GetFileName(file.Path);
```
chuckries, line 241: @tmat Mac tests don't like this because the test input is windows style paths and I'm assuming Path.GetFileName(...) looks for forward slashes on that OS.

should this just be something like
```cs
                    int lastSeparatorIndex = file.Path.LastIndexOfAny(new[] { '\\', '/' });
                    relativeFilePath = lastSeparatorIndex > 0 ? file.Path.Substring(lastSeparatorIndex + 1) : file.Path;
```

tmat: Yes, can't use GetFileName.

src/SourceLink.Tools/SourceLinkMap.cs
```diff
@@ -189,6 +189,24 @@ private static bool TryParseEntry(string key, string value, out Entry entry)
...
             out string? uri)
+        {
+            return TryGetUri(path, out uri, out _);
+        }
+
+        /// <summary>
+        /// Maps specified <paramref name="path"/> to the corresponding URL.
+        /// </summary>
+        /// <exception cref="ArgumentNullException"><paramref name="path"/> is null.</exception>
+        public bool TryGetUri(
```
tmat, line 200: Add doc comment for `relativeFilePath`

src/SourceLink.Tools/SourceLinkMap.cs
tmat: Can we set `relativeFilePath = null` here? Since there is no relative path... the caller can use the file name, if they want to.

PR #711, opened 2021-03-29
Migrate to OneLocBuild

jonfortescue:
Documentation (pending merge into arcade) can be found [here](https://github.com/jonfortescue/arcade/blob/OneLocBuildDoc/Documentation/OneLocBuild.md).

Sample of this task working in a build [here](https://dev.azure.com/dnceng/internal/_build/results?buildId=1048311&view=results).

Included files from eng/common because sourcelink is currently blocked on arcade merge. There will be no merge conflicts with latest arcade.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -31,6 +31,12 @@ stages:
 - stage: build
   displayName: Build
   jobs:
+  - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+    - template: /eng/common/templates/job/onelocbuild.yml
+      parameters:
+        CreatePr: false
+        LclSource: lclFilesfromPackage
+        LclPackageId: 'LCL-JUNO-PROD-DOTNETSDK'
   - template: /eng/common/templates/jobs/jobs.yml
     parameters:
       enableMicrobuild: true
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -36,7 +36,7 @@ stages:
       parameters:
         CreatePr: false
         LclSource: lclFilesfromPackage
-        LclPackageId: 'LCL-JUNO-PROD-DOTNETSDK'
+        LclPackageId: 'LCL-JUNO-PROD-SOURCELINK'
   - template: /eng/common/templates/jobs/jobs.yml
     parameters:
       enableMicrobuild: true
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -31,6 +31,12 @@ stages:
 - stage: build
   displayName: Build
   jobs:
+  - ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
+    - template: /eng/common/templates/job/onelocbuild.yml
+      parameters:
+        CreatePr: false
+        LclSource: lclFilesfromPackage
+        LclPackageId: 'LCL-JUNO-PROD-SOURCELINK'
```
tmat, line 39: Why is this not part of the common template `jobs.yml`?

tmat:
Is the script reading from `$(RepoRoot)/Localize/LocExclusions.json` and writing to the same directory? I don't think that's ideal. I'd prefer the configuration be in `$(RepoRoot)/eng/common` directory and then the output was in `(RepoRoot)/artifacts/localization` directory. Can we change that?

azure-pipelines.yml
jonfortescue: There are a lot of repos that need it that don't use `jobs.yml` and a lot of repos that use `jobs.yml` that don't need it. It added a bunch more parameters to `jobs.yml` that just got plumbed through to this template so we ultimately decided to just keep it separate.

jonfortescue:
@tmat That's something we can do but because it's part of eng/common it's best to not do it as part of this PR. I've filed https://github.com/dotnet/core-eng/issues/12709 to track it.

azure-pipelines.yml
tmat: I see only one parameter here that'd need to be passed: the LclPackageId. that also is cryptic to me and I'd think it could be synthesized based on the name of the repo. So really, I think we just need a single parameter: `enableLocalization: true` similar to other `enable*` parameters.

jonfortescue: Here, that's true. However, across all the repos that use this there's a whole host of different configurations that need to be passed. What's more, the package IDs can unfortunately not be synthesized from the repo name (e.g. deployment-tools where the package ID is `LCL-JUNO-PROD-MAGE`).

We could look at making use of `jobs.yml` in the next iteration (which there is funding for and will start immediately), but simpleloc is being turned off today and pushing out that change will require waiting for an arcade promotion.

tmat: Following up on it as a separate effort sounds good. I wonder what the other parameters are and if we could avoid them. Also is it possible to change the PackageId to match the repo name - if only to make it easier to decipher what they mean.

tmat: Thanks!

PR #714, opened 2021-04-27
Add localization from OneLocBuild (2021-04-27)

jonfortescue:


jonfortescue:
Best way to review these changes is to turn off whitespace changes when reviewing.

PR #719, opened 2021-07-02
Turn on localization PRs from OneLocBuild

jonfortescue:
Documentation on this change [here](https://github.com/dotnet/arcade/blob/main/Documentation/OneLocBuild.md).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -29,10 +29,10 @@ stages:
 - stage: build
   displayName: Build
   jobs:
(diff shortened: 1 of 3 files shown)


tmat:
I have merged Arcade updates, so we should have the latest now. @jonfortescue can you confirm this is no longer needed?

jonfortescue:
I'll still need to add the mirror repo, but the manual changes arcade updates are unnecessary now. Will update these PRs accordingly.

PR #742, opened 2021-08-27
Conditionally build for only 6.0 when building from source

### LINKED ISSUE #724: Update ArPow build to build net6.0 TFM
When building source-build, all sources are built with the current TFM.  Because of this, patches were introduced to build sourcelink with the net6.0 TFM.  The patches are:
- https://github.com/dotnet/installer/blob/main/src/SourceBuild/tarball/content/patches/sourcelink/0001-Update-TFMs-to-net6.0.patch
- https://github.com/dotnet/installer/blob/main/src/SourceBuild/tarball/content/patches/sourcelink/0002-Update-Task-paths-to-include-net6.0.patch

These patches should be included in the sourcelink source so they are built in the SourceBuild leg.

dseefeld:
Fixes #724


tmat (approved):
I'm still not quite sure why we need to override TargetFrameworks and remove net472 in each project when building from source, but will approve to unblock.

PR #774, opened 2021-11-30
Mark SourceLinkMap as auto-generated

davidwengier:
Because this is a source package, including it in other projects can be problematic due to their own code style rules, namely Roslyn CI builds fail because of whitespace issues and the file header being different.

If there is a better solution than this I'm all for it, just thought I'd put this up to not lose track of the issue. I've worked around it for now by copying the source into Roslyn.


tmat (changes requested):
This disables code style checking in this repository as well. 

Instead, Roslyn analyzers should treat files imported from source packages as generated,

PR #776, opened 2021-12-10
Update LibGit to 0.26.2

tmat:


bording:
@tmat I recommend using the latest 0.27 preview instead. Lots of progress on native binaries that get shipped to increase compatibility.

tmat:
We would like to use a stable version. It's just for testing.

tmat:
Superseded by https://github.com/dotnet/sourcelink/pull/778.

@bording
Turns out 0.27 preview was needed to run on the latest Ubuntu image we have.

PR #933, opened 2023-01-02
add pre-build detection

oleksandr-didyk:
Resolves https://github.com/dotnet/sourcelink/issues/932

Additionally created two subscriptions for the dependencies on `runtime` and `msbuild` (disabled until this PR is merged):
```
https://github.com/dotnet/msbuild (VS 17.5) ==> 'https://github.com/dotnet/sourcelink' ('main')
  - Id: 091aef12-b872-444d-b223-08daecc804bf
  - Update Frequency: EveryWeek
  - Enabled: False
  - Batchable: False
  - PR Failure Notification tags:
  - Merge Policies: []
https://github.com/dotnet/runtime (.NET 8) ==> 'https://github.com/dotnet/sourcelink' ('main')
  - Id: e02204af-e240-492f-2682-08daecc8f206
  - Update Frequency: EveryWeek
  - Enabled: False
  - Batchable: False
  - PR Failure Notification tags:
  - Merge Policies: []
```


NuGet.config
```diff
@@ -5,6 +5,8 @@
     <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
     <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
+    <add key="dotnet8-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8-transport/nuget/v3/index.json" />
+    <add key="dotnet8" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" />
```
oleksandr-didyk, line 9: Added for `runtime` dependency

azure-pipelines.yml
```diff
@@ -145,7 +145,7 @@ stages:
           steps:
           - checkout: self
             clean: true
-          - script: eng/cibuild.sh --configuration $(_BuildConfig) --prepareMachine
+          - script: eng/cibuild.sh --configuration $(_BuildConfig) --prepareMachine /p:SourceBuildIntermediateNupkgRid='linux-x64'
```
oleksandr-didyk, line 148: Added since `runtime`'s intermediate is not `ManagedOnly`

src/SourceLink.Tools/Microsoft.SourceLink.Tools.Package.csproj
```diff
@@ -6,6 +6,7 @@
     <GenerateDocumentationFile>false</GenerateDocumentationFile>
     <DebugType>none</DebugType>
     <GenerateDependencyFile>false</GenerateDependencyFile>
+    <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings> <!-- For System.Text.Json 8.x.y reference -->
```
oleksandr-didyk, line 9: Saw other repositories do this, not sure if this is the ideal solution here @MichaelSimons

tmat: Please add a comment describing what the purpose of the property is.

azure-pipelines.yml
tmat: What does this property do and why isn't it set in `common/cibuild.sh`?

eng/Versions.props
```diff
@@ -9,17 +9,19 @@
     <!-- Opt-in repo features -->
     <UsingToolNetFrameworkReferenceAssemblies>true</UsingToolNetFrameworkReferenceAssemblies>
     <UsingToolSymbolUploader>true</UsingToolSymbolUploader>
-    <MicrosoftBuildVersion>16.7.0</MicrosoftBuildVersion>
-    <MicrosoftBuildTasksCore>16.7.0</MicrosoftBuildTasksCore>
+    <!-- latest versions -->
```
tmat, line 12: Are these versions auto updated by Arcade? If so, please mention it in the comment.

src/SourceLink.Tools/Microsoft.SourceLink.Tools.Package.csproj
ViktorHofer: This property shouldn't be set without a VERY strong reason. It's a hack to workaround a repository targeting a lower .NETCoreApp TFM than what the referenced packages actually support.

ViktorHofer:
```suggestion
```

ViktorHofer: If you want to reference the live version of System.Text.Json then this project must be updated to net8.0 as source-build packages only include the latest .NETCoreApp TFM: net8.0. Falling back to the netstandard2.0 TFM isn't a great workaround either as that behaves differently than the net8.0 asset.

azure-pipelines.yml
oleksandr-didyk: The property specifies to source-build MSBuild logic the RID of an intermediate package. In cases where intermediate packages are platform-specific (like with `runtime`), this property has to be set or the source-build will fail to `restore`. In this case, we are restoring a portable intermediate, hence the `linux-x64` value.

My bad, I did not notice there was a different script in `eng`, will put it there instead.

As for `common/cibuild.sh` - it is coming from Arcade, so any change there would get overridden / conflict once a new Arcade version flows in, no?

eng/Versions.props
oleksandr-didyk: They are updated by Darc automatically, yes. Of course, will clarify this in the comment

azure-pipelines.yml
tmat: Why is this property needed for this repo and not other repos?

eng/Version.Details.xml
```diff
@@ -1,16 +1,41 @@
...
+      <Sha>a6f6699d1f70bf79db82030938d2c5e52d1e4d2e</Sha>
+    </Dependency>
   </ProductDependencies>
   <ToolsetDependencies>
     <Dependency Name="Microsoft.DotNet.Arcade.Sdk" Version="8.0.0-beta.22630.1">
       <Uri>https://github.com/dotnet/arcade</Uri>
       <Sha>e82404fca08383513e0b0b3c5308d4a9b18b7c7a</Sha>
+      <SourceBuild RepoName="arcade" ManagedOnly="true" />
     </Dependency>
     <Dependency Name="Microsoft.SourceBuild.Intermediate.source-build-reference-packages" Version="8.0.0-alpha.1.23055.1">
```
MichaelSimons, line 25: Since you are touching this file, can you make this a ProductDependency?

oleksandr-didyk: Yeah, thanks for the reminder, will change

azure-pipelines.yml
oleksandr-didyk: No other repository that I have worked with was using a runtime intermediate. We are on our way to enable pre-built detection in repos, so most probably others that are dependent on latest runtime would get a similar property

tmat: Pardon my ignorance, but what does that mean that the "runtime intermediate" is being used? How does that happen?

eng/cibuild.sh
```diff
@@ -13,4 +13,5 @@ while [[ -h $source ]]; do
 done
 scriptroot="$( cd -P "$( dirname "$source" )" && pwd )"
 
-. "$scriptroot/common/build.sh" --restore --build --test --pack --integrationTest --publish --ci $@
+# SourceBuildIntermediateNupkgRid indicates to source-build that portable intermediates should be restored (required for runtime dependency)
+. "$scriptroot/common/build.sh" --restore --build --test --pack --integrationTest --publish --ci /p:SourceBuildIntermediateNupkgRid="linux-x64" "$@"
```
MichaelSimons, line 17: Having to specify /p:SourceBuildIntermediateNupkgRid="linux-x64" does not seem like the optimal e2e UX here.  Can't the ArPOW infrastructure infer this?

azure-pipelines.yml
oleksandr-didyk: Intermediate packages are a transport format we use to transfer source-built pre-builts between repositories ([docs](https://github.com/dotnet/source-build/blob/main/Documentation/planning/arcade-powered-source-build/intermediate-nupkg.md)).

In this case, during SourceLink's source-build the build logic will retrieve the runtime's source-build by restoring the intermediate packages and use assets from there in the build. These assets can be platform-specific, hence there are different packages for different platforms. For the build logic to know which exact intermediate to pull, we need to specify the `SourceBuildIntermediateNupkgRid` property

eng/cibuild.sh
oleksandr-didyk: Have relatively little experience with ArPOW, but sounds doable. It would take some time for it to flow in though, right?

MichaelSimons: Per a discussion offline, @oleksandr-didyk is going to add infra into arcade to handle this so that repo's don't have to specify this.

azure-pipelines.yml
tmat: You're saying:
> No other repository that I have worked with was using a runtime intermediate. 

I'm still not understanding what's special about Source Link build that would be different from any other repository that builds a dotnet library.

tmat: Answered here: https://github.com/dotnet/sourcelink/pull/933#discussion_r1068242626

eng/cibuild.sh
oleksandr-didyk: The functionality was added to the arcade infra but it currently cannot be flown-in due to new Arcade version bumping the .NET version to 8.0. Mentioned this in the relevant Maestro PR -> https://github.com/dotnet/sourcelink/pull/936#issuecomment-1411884460

mthalman:
@oleksandr-didyk - Can you give a status on this please? Is there in-progress work or is it blocked or just pending work? Thanks

oleksandr-didyk:
> @oleksandr-didyk - Can you give a status on this please? Is there in-progress work or is it blocked or just pending work? Thanks

It was initially waiting for the Arcade bump to flow in. The change got into `main` last week so this PR got unblocked.
I updated it and it should be ready for re-review now

azure-pipelines.yml
oleksandr-didyk: Removed with the new changes from Arcade (SB build logic will identify the RID param by itself)

src/SourceLink.Tools/Microsoft.SourceLink.Tools.Package.csproj
oleksandr-didyk: Removed since the projects now utilize `$(NetCurrent)` (`net8.0`)

eng/cibuild.sh
oleksandr-didyk: Removed with the new changes from Arcade (SB build logic will identify the RID param by itself)

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -1,5 +1,11 @@
+<!-- See aka.ms/dotnet/prebuilts for guidance on what pre-builts are and how to eliminate them. -->
 <UsageData>
   <IgnorePatterns>
-    <UsagePattern IdentityGlob="*/*" />
+    <UsagePattern IdentityGlob="Microsoft.SourceBuild.Intermediate.*" />
   </IgnorePatterns>
-</UsageData>
+  <Usages>
+    <!-- MSBuild versions starting from 16.10.0 do not target NET Core -->
+    <!-- and fail the non-source-build -->
```
oleksandr-didyk, line 8: The exception in the main build is the following:

```
error NU1701: Package 'Microsoft.Build 16.10.0' was restored using '.NETFramework,Version=v4.6.1, ..., .NETFramework,Version=v4.8.1' instead of the project target framework '.NETCoreApp,Version=v3.1'. 
This package may not be fully compatible with your project.
```

eng/Version.Details.xml
```diff
@@ -13,16 +13,44 @@
       <Uri>https://github.com/dotnet/command-line-api</Uri>
       <Sha>209b724a3c843253d3071e8348c353b297b0b8b5</Sha>
     </Dependency>
+    <Dependency Name="Microsoft.SourceBuild.Intermediate.command-line-api" Version="0.1.327201">
+      <Uri>https://github.com/dotnet/command-line-api</Uri>
+      <Sha>209b724a3c843253d3071e8348c353b297b0b8b5</Sha>
+    </Dependency>
```
MichaelSimons, line 19: Shouldn't this be marked as source-build?

eng/SourceBuildPrebuiltBaseline.xml
```diff
@@ -1,5 +1,11 @@
...
+    <Usage Id="Microsoft.Build" Version="16.8.0" IsDirectDependency="true" />
```
MichaelSimons, line 9: Should an SBRP be used here instead?

oleksandr-didyk: SBRP has 16.10+ or 15.7 versions, 16.10 cannot be used per comment + did not want to downgrade a major version.

Can add the 16.8 or 16.9 back to SBRP and then remove this exclusion (wasn't sure, wanted to discuss this beforehand)

MichaelSimons: Repo owners can comment on the specific version needed.  The exclusion here will need to be addressed otherwise this will be reported as a prebuilt in the product build once the per repo pvp feature is enabled because it will force the product build to use the version specified in this repo since the dependency is not marked as source-build.

oleksandr-didyk: @tmat the version currently used is 16.7.0; does it makes sense from your perspective to keep using this version?

eng/Versions.props
```diff
@@ -8,18 +8,19 @@
     <!-- Opt-in repo features -->
     <UsingToolNetFrameworkReferenceAssemblies>true</UsingToolNetFrameworkReferenceAssemblies>
     <UsingToolSymbolUploader>true</UsingToolSymbolUploader>
-    <MicrosoftBuildVersion>16.7.0</MicrosoftBuildVersion>
-    <MicrosoftBuildTasksCore>16.7.0</MicrosoftBuildTasksCore>
+    <!-- latest versions auto-updated by Maestro/Darc tooling -->
+    <SystemTextJsonVersion>8.0.0-preview.4.23210.22</SystemTextJsonVersion>
+    <!-- Pinned versions -->
```
tmat, line 13: Should we flow the latest msbuild version as well?

eng/SourceBuildPrebuiltBaseline.xml
tmat: We can use any msbuild version that works. The repo defines msbuild tasks, not using any special msbuild features.

eng/Versions.props
oleksandr-didyk: We technically can't due to MSBuild versions after 16.10 not targeting `netcoreapp3.1` (exception at https://github.com/dotnet/sourcelink/pull/933#discussion_r1162691841). And as far as I am aware of we don't want the two builds diverge, as in we would not want them to use different versions of dependencies

ViktorHofer: The latest officially released MSBuild package supports net7.0. Why can't we use that TFM? Sourcelink's source build is about building the very latest. Why do we care about netcoreapp3.1?

oleksandr-didyk: Wouldn't we need then 2 MSBuild versions - one for source-build and one for the default build?

rainersigwald: It should be possible through the `netstandard2.0` ref assemblies we package, yes. Works on my trivial test library . . .

rainersigwald: Correct, refs to build tasks should be `ExcludeAssets="runtime" PrivateAssets="all"`.

ViktorHofer: > Correct, refs to build tasks should be ExcludeAssets="runtime" PrivateAssets="all".

@rainersigwald thanks. Last question just to double check, .NET Framework compiled tasks should bundle the MSBuild assemblies with it, right?

@oleksandr-didyk based on the above, we should be able to use the latest MSBuild packages just fine.

rainersigwald: Don't bundle MSBuild assemblies in any TF--any time a task is run, MSBuild will force the use of its own assemblies so they never need to be redistributed.

ViktorHofer: While looking into this more closely I noticed that sourcelink depends on the "Microsoft.Build" package because they us evaluation helper functions which aren't available in the other packages: https://github.com/dotnet/sourcelink/blob/40a5327a26ef250235be54143138325aedd35b79/src/Microsoft.Build.Tasks.Git/GitOperations.cs#L267

That's why libraries need to upgrade from netcoreapp3.1 to net6.0 to use the latest available msbuild package in SBRP: 17.3.2. I submitted https://github.com/dotnet/sourcelink/pull/1001 to fix that.

ViktorHofer: FWIW, sourcelink can't go beyond 17.3.2 because the latest Microsoft.Build doesn't support net6.0 anymore. Even though I chatted with Rainer about this in other threads some time ago, msbuild's package TFM support matrix still confuses me. Does the API that sourcelink uses (`ProjectCollection.Escape`) justify the locked dependency?

oleksandr-didyk: > I submitted https://github.com/dotnet/sourcelink/pull/1001 to fix that.

Thanks! I will wait for that bump to flow then, unless a different decision around MSBuild's version arises in the meantime

ViktorHofer: @oleksandr-didyk https://github.com/dotnet/sourcelink/pull/1001 is merged and this PR should now be unblocked.

NuGet.config
```diff
@@ -6,6 +6,9 @@
     <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
     <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
     <add key="dotnet8" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" />
+    <!-- Feeds for source-build intermediates -->
+    <add key="dotnet-libraries-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-libraries-transport/nuget/v3/index.json" />
+    <add key="dotnet8-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8-transport/nuget/v3/index.json" />
```
ViktorHofer, line 11: nit: please move dotnet8-transport after dotnet8. That's the order that we use in other repositories.

eng/SourceBuildPrebuiltBaseline.xml
ViktorHofer, line 9: Remove this entry now.

ViktorHofer:
OK there's one more PR that I just submitted that will again make things easier and this PRs diff much smaller: https://github.com/dotnet/sourcelink/pull/1003. I suggest to also wait for that one.

eng/SourceBuildPrebuiltBaseline.xml
oleksandr-didyk: With changes in https://github.com/dotnet/sourcelink/pull/1001 we can use 17.3.2 which already has an SBRP for it, removed the exclusion

NuGet.config
oleksandr-didyk: Sure thing, will do

MichaelSimons: I see this still exists in the latest changes.  Can this be removed?

src/dotnet-sourcelink/dotnet-sourcelink.csproj
```diff
@@ -4,6 +4,7 @@
     <TargetFrameworks>$(NetCurrent)</TargetFrameworks>
     <!-- Allow tool to roll forward to a newer major version. -->
     <RollForward>Major</RollForward>
+    <ExcludeFromSourceBuild>true</ExcludeFromSourceBuild>
```
ViktorHofer, line 7: Why does this tool need to be excluded from source build? Please add a comment as it isn't clear just from reading the code.

tmat: It should be included. This is not a shipping product yet but it is meant to be. It's for customers to test that their binaries have correct Source Link.

oleksandr-didyk: >This is not a shipping product yet

Can this be done as a separate issue to not block this PR? Or it being excluded for some time is a no-go?

ViktorHofer: IMO if source build repositories currently don't depend on the tool, it shouldn't be an issue to exclude it for now and track enabling it.

oleksandr-didyk: Created an issue for it -> https://github.com/dotnet/sourcelink/issues/1028

tmat (approved):

PR #988, opened 2023-04-05
Prepare for inclusion in .NET SDK

tmat:


tmat:
@KalleOlaviNiemitalo Standalone packages will be still available. The one in SDK will be suppressed if an explicit SourceLink PackageReference is used in the project.

PR #1001, opened 2023-04-15
Upgrade to net6.0 and remove unused dependencies

ViktorHofer:
1. Upgrade netcoreapp3.1 to net6.0 as the former is out-of-support and not supported by MSBuild/17.3.2
2. Remove unused package versions
3. Upgrade System.Text.Json to the latest available version (in SBRP and nuget.org) 7.0.2.
4. Directly reference NETStandard.Library/2.0.3 in test projects to upgrade the transitive NETStandard.Library/1.6.1 version (which comes from xunit) as 1.6.1 is vulnerable and raises CG warnings.

Contributes to https://github.com/dotnet/sourcelink/pull/933

cc @MichaelSimons (removes netstandard1.x dependencies), @oleksandr-didyk

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -6,20 +6,20 @@
     <VersionPrefix>8.0.0</VersionPrefix>
     <PreReleaseVersionLabel>beta</PreReleaseVersionLabel>
     <!-- Opt-in repo features -->
-    <UsingToolNetFrameworkReferenceAssemblies>true</UsingToolNetFrameworkReferenceAssemblies>
     <UsingToolSymbolUploader>true</UsingToolSymbolUploader>
-    <MicrosoftBuildVersion>16.7.0</MicrosoftBuildVersion>
-    <MicrosoftBuildTasksCore>16.7.0</MicrosoftBuildTasksCore>
+    <MicrosoftBuildVersion>17.3.2</MicrosoftBuildVersion>
+    <MicrosoftBuildTasksCore>17.3.2</MicrosoftBuildTasksCore>
+
+    <!-- TODO: Remove after https://github.com/dotnet/arcade/pull/13178 is consumed. -->
+    <MicrosoftNETTestSdkVersion>17.5.0</MicrosoftNETTestSdkVersion>
+
     <MicrosoftTeamFoundationServerExtendedClientVersion>19.210.0-preview</MicrosoftTeamFoundationServerExtendedClientVersion>
-    <MicrosoftDotNetPlatformAbstractionsVersion>3.1.6</MicrosoftDotNetPlatformAbstractionsVersion>
     <NuGetVersioningVersion>5.7.0</NuGetVersioningVersion>
     <NuGetPackagingVersion>5.7.0</NuGetPackagingVersion>
     <SystemCommandLineVersion>2.0.0-beta4.22272.1</SystemCommandLineVersion>
     <SystemCommandLineNamingConventionBinderVersion>2.0.0-beta4.22272.1</SystemCommandLineNamingConventionBinderVersion>
     <SystemCommandLineRenderingVersion>0.4.0-alpha.22272.1</SystemCommandLineRenderingVersion>
-    <SystemReflectionMetadataVersion>1.8.1</SystemReflectionMetadataVersion>
-    <SystemValueTupleVersion>4.5.0</SystemValueTupleVersion>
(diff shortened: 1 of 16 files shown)


eng/Versions.props
```diff
@@ -6,20 +6,20 @@
...
-    <MicrosoftDotNetPlatformAbstractionsVersion>3.1.6</MicrosoftDotNetPlatformAbstractionsVersion>
     <NuGetVersioningVersion>5.7.0</NuGetVersioningVersion>
     <NuGetPackagingVersion>5.7.0</NuGetPackagingVersion>
     <SystemCommandLineVersion>2.0.0-beta4.22272.1</SystemCommandLineVersion>
     <SystemCommandLineNamingConventionBinderVersion>2.0.0-beta4.22272.1</SystemCommandLineNamingConventionBinderVersion>
     <SystemCommandLineRenderingVersion>0.4.0-alpha.22272.1</SystemCommandLineRenderingVersion>
-    <SystemReflectionMetadataVersion>1.8.1</SystemReflectionMetadataVersion>
-    <SystemValueTupleVersion>4.5.0</SystemValueTupleVersion>
-    <SystemTextJsonVersion>4.7.2</SystemTextJsonVersion>
+    <SystemTextJsonVersion>7.0.2</SystemTextJsonVersion>
```
oleksandr-didyk, line 22: Couldn't `sourcelink` consume the latest (`net8`) version of `System.Text.Json`?

oleksandr-didyk (approved):

eng/Versions.props
ViktorHofer: Potentially yes. I picked the latest available version from SBRP. Feel free to update to the "live" version in https://github.com/dotnet/sourcelink/pull/933

ViktorHofer: Do you mean the live version? I don't think that's possible as System.Text.Json builds in dotnet/runtime which builds after dotnet/sourcelink.

ViktorHofer: cc @mmitche

MichaelSimons: It can use the "live" version but it technically would be the n-1 version from the previously source-built artifacts because of @ViktorHofer notes, sourcelink builds before runtime.  In this case it feels like an SBRP version feels like the better choice.

mmitche: I think SBRP is the right choice here.

mmitche: It's also not going to depend on new surface area, so building against 7.0.x and deploying/running previously source built is  simpler.

src/Directory.Build.targets
```diff
@@ -3,10 +3,15 @@
 <Project>
   <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
 
+  <ItemGroup Condition="'$(IsTestProject)' == 'true'">
+    <!-- Upgrade the NETStandard.Library transitive xunit dependency to avoid transitive 1.x NS dependencies. -->
```
tmat, line 7: > [](http://example.com/codeflow?start=4&length=109)

Is this just a temporary workaround?

tmat:
> <?xml version="1.0" encoding="utf-8"?>

Licensing header removed?

---
Refers to: src/Microsoft.Build.Tasks.Git.UnitTests/Microsoft.Build.Tasks.Git.UnitTests.csproj:1 in e70957e. [](commit_id = e70957e005b0e1bfd7c5db01b9a5ec26f5e51d5a, deletion_comment = True)

ViktorHofer:
> Licensing header removed?

Licensing headers are only required in shipping artifacts like source code or msbuild props and targets files. Project files aren't shipping assets and hence don't need these licensing headers. @jkotas let me know that some years ago. That's why we don't have in the project files in dotnet/runtime. Most other repositories don't have them either.

src/Directory.Build.targets
ViktorHofer: We discussed adding this infrastructure to the SDK as it minimizes the dependency graph. Back then we decided not add it into the SDK just for the sake of risk but this has been in runtime and other repositories for years: https://github.com/dotnet/runtime/blob/048467b13160399d493b97a90b2117f58f01ac51/eng/testing/xunit/xunit.targets#L3-L6

It minimizes the dependency graph by not bringing in the NETSTandard.Library/1.6.1 transitive dependeny. We can remove this when xunit decides to publish a new version with a netstandard2.0 or modern .NETCoreApp asset.

tmat (approved):

PR #1003, opened 2023-04-18
Clean-up sourcelink repo and remove workarounds

ViktorHofer:
1. ~~Remove license headers from non-shipping assets (project files)~~ https://github.com/dotnet/sourcelink/pull/1014
2. Use the Microsoft.Build.NoTargets SDK for content-only packages That avoids invoking the compiler without defining custom targets and makes the repository use the same path as other repositories in the stack.
3. Delete runtimeconfig.template.json in favor of the in-built `"<RollForward>...</RollForward>"` switch.
4. Remove not needed sourcebuild target.
5. Group dependencies in Versions.props by repositories. Remove the now unneeded Microsoft.NET.Test.Sdk entry.
6. Define target frameworks for source build centrally in one place.
7. Remove unnecessary package dependencies from projects.
8. Apply code styling:
   - "empty line after Project tag and before closing Project tag"
   - "empty line between groups (property, item, target)"
   - "TargetFramework(s) should be the first property in the project"
9. Remove unnecessary PrivateAssets="all" attributes from four ProjectReference items.

PR #1015, opened 2023-04-27
Clean-up sourcelink repo and remove workarounds (grouped into commits)

ViktorHofer:
Replaces https://github.com/dotnet/sourcelink/pull/1003

**Grouped into commits**

I will send separate PRs after this one is in for the formatting changes and updating the TFM to use `$(NetFrameworkToolCurrent)` (which requires a new Arcade update).

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -1,7 +1,7 @@
 <Project>
   <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
 
-  <!-- TODO: Remove when Arcade offers an in-built way to filter out anything other than NetCurrent. -->
+  <!-- TODO: Remove when Arcade offers an in-built way to filter out anything other than NetCurrent: https://github.com/dotnet/arcade/issues/13390. -->
   <PropertyGroup>
     <TargetFrameworks Condition="'$(TargetFrameworks)' != '' and '$(DotNetBuildFromSource)' == 'true'">$(NetCurrent)</TargetFrameworks>
   </PropertyGroup>
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -1,6 +1,11 @@
 <Project>
   <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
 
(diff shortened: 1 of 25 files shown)


ViktorHofer:
Resolved the merge conflict. cc @tmat

src/Directory.Build.targets
```diff
@@ -1,6 +1,11 @@
 <Project>
   <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
 
+  <!-- TODO: Remove when Arcade offers an in-built way to filter out anything other than NetCurrent. -->
```
tmat, line 4: Please, include a link to the issue tracking this work.

src/Microsoft.Build.Tasks.Git.UnitTests/Microsoft.Build.Tasks.Git.UnitTests.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Xunit.Combinatorial" Version="$(XunitCombinatorialVersion)" />
```
tmat, line 8: I'd rather haver this references in the TestUtilities so that it is available to all tests. What's the benefit of moving it here?

ViktorHofer: Having fine grained dependencies make restore more efficient and reduces package dependency graph which is emitted into both `project.assets.json` and `<app>.deps.json`.

That said, I don't have a strong preference here at all. If you want this to remain a transitive dependencies, I'm fine with that as well.

tmat: I'd prefer simplicity over performance here.

ViktorHofer: Reverted the change

tmat (approved):

PR #1078, opened 2023-08-01
Do not report warning if submodule dir is missing

tmat:
Fixes https://github.com/dotnet/sourcelink/issues/1071


AArnott:
@tmat, my submodule isn't missing. It's very much checked out. But [it still complains](https://github.com/AArnott/Nerdbank.Cryptocurrencies/actions/runs/5912067208/job/16035286721#step:6:16). 
This is on .NET SDK 7.0.302. Can I expect your fix here will also address my issue?

AArnott:
Another issue I'm having is that the warning doesn't include a diagnostic code, so I can't suppress the warning. Since I have warnings as errors in my CI/PR builds, this is a blocking issue.

PR #1119, opened 2023-10-27
Use csproj as pack input instead of nuspec

ViktorHofer:
Now that sourcelink is a native part of the .NET SDK, Arcade doesn't bring sourcelink packages in anymore. Now that the cyclic package dependency is avoided, these projects don't need to use nuspecs anymore.

This removes custom infrastructure and hacks, and allows better source build controls.

I diffed the produced packages and the content is identical (a deps.json file is added but that's recommended by msbuild for build tasks these days).


ViktorHofer:
> 
Does the team intend to keep publishing sourcelink NuGet packages for use with older SDKs?

Yes and this PR doesn't change that. The .NET 9 source link packages will work on all the officially supported MSBuild flavours.

ViktorHofer:
Unfortunately I can't reproduce the failures locally: `Microsoft.SourceLink.Git.IntegrationTests`.

ViktorHofer:
cc @tmat

jaredpar:
@davidwengier can you take a look at this?

src/Microsoft.Build.Tasks.Git/build/Microsoft.Build.Tasks.Git.props
```diff
@@ -2,6 +2,7 @@
 <Project>
   <PropertyGroup>
     <MicrosoftBuildTasksGitAssemblyFile Condition="'$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)..\tools\net472\Microsoft.Build.Tasks.Git.dll</MicrosoftBuildTasksGitAssemblyFile>
-    <MicrosoftBuildTasksGitAssemblyFile Condition="'$(MSBuildRuntimeType)' == 'Core'">$(MSBuildThisFileDirectory)..\tools\core\Microsoft.Build.Tasks.Git.dll</MicrosoftBuildTasksGitAssemblyFile>
+    <MicrosoftBuildTasksGitAssemblyFile Condition="'$(MSBuildRuntimeType)' == 'Core' and '$(MSBuildVersion)' &lt; '17.8.0'">$(MSBuildThisFileDirectory)..\tools\net6.0\Microsoft.Build.Tasks.Git.dll</MicrosoftBuildTasksGitAssemblyFile>
```
tmat, line 5: Why not keep the directory name "core" so that we don't need to add (and maintain) these extra conditions?

eng/BuildTask.targets
```diff
@@ -0,0 +1,97 @@
...
+    <PackageReference Update="NuGet.Commands" Publish="false" PrivateAssets="all" />
+    <PackageReference Update="NuGet.Common" Publish="false" PrivateAssets="all" />
+    <PackageReference Update="NuGet.Configuration" Publish="false" PrivateAssets="all" />
+    <PackageReference Update="NuGet.Frameworks" Publish="false" PrivateAssets="all" />
+    <PackageReference Update="NuGet.Packaging" Publish="false" PrivateAssets="all" />
+    <PackageReference Update="NuGet.ProjectModel" Publish="false" PrivateAssets="all" />
+    <PackageReference Update="NuGet.Versioning" Publish="false" PrivateAssets="all" />
+  </ItemGroup>
+
+  <!-- Don't include assemblies that are inbox in Desktop MSBuild -->
```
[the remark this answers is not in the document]
ViktorHofer: As already mentioned before, no. The minimum supported version is still the one that can load net6.0 task assemblies.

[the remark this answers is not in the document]
jaredpar: The extended support of VS doesn't necessarily mean that we will provide new features and functionality. It's about applying security and critical fixes. If such a need arises for source link here we could provide ti by forking off a previous version and patching that. It shouldn't stop our progress from depending on newer MSBuild features for the lifetime of the extended support.

src/Microsoft.Build.Tasks.Git/build/Microsoft.Build.Tasks.Git.props
ViktorHofer: Grouping assemblies into a TFM subfolder is the pattern that we follow in other tools packages (i.e. Arcade) and aligns with NuGet's conventions for lib/ref folders.

Source-build requires to target the latest TFM and back-compatibility with older SDKs suggests to target the minimum supported tool TFM (net6.0 at the moment). To align with these goals, we recommend to build the super-set of TFMs in the regular build and make source build only build the latest TFM graph slice. 

Now that projects use floating TFM properties (`NetToolCurrent`, `NetToolMinimum` and `NetFrameworkCurrent`) the remaining places that need a manual TFM update (every year) are these hardcoded values in props and targets files.

I opened https://github.com/dotnet/arcade/issues/14138 to track adding infrastructure to Arcade to avoid these hardcodes via template substitution.

tmat: > Grouping assemblies into a TFM subfolder is the pattern that we follow in other tools packages (i.e. Arcade) and aligns with NuGet's conventions for lib/ref folders.

I don't see the value of it for build tasks targets though since we are not multi-targeting them. We only ever need a one target for desktop and one for core, no?

Not using the TFMs would reduce complexity.

tmat (changes requested):
We should avoid including the TFM in the `tools` package path. It makes things unnecessarily complicated.

ViktorHofer:
I finally found time to continue to work on this. @tmat, please take a look. I used the same approach that we came up with in arcade for build task projects. Btw, it would be good to switch this repository to Central Package Management with a future change.

tmat:
> The .NET Framework output is significantly different as it now includes all dependencies that aren't supplied by the MSBuild inside Visual Studio.

Would you mind listing the differences?

eng/BuildTask.targets
```diff
@@ -0,0 +1,118 @@
...
+    <PackageReference Update="Microsoft.Extensions.DependencyModel" Version="7.0.0" />
+    <PackageReference Update="System.Collections.Immutable" Version="7.0.0" />
+    <PackageReference Update="System.Reflection.Metadata" Version="7.0.0" />
+  </ItemGroup>
+
+  <!-- Publish .NET Core assets and include them in the package under $(BuildTaskTargetFolder) directory. -->
+  <Target Name="_AddBuildOutputToPackageCore" DependsOnTargets="Publish" Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp'">
+    <PropertyGroup>
+      <BuildTaskTargetTfmSpecificFolder Condition="'$(BuildTaskTargetTfmSpecificFolder)' == ''">net</BuildTaskTargetTfmSpecificFolder>
+    </PropertyGroup>
```
tmat, line 101: We shouldn't need this here, let's hard-code "net" and not make it customizable.

src/Microsoft.Build.Tasks.Git/buildTransitive/Microsoft.Build.Tasks.Git.props
```diff
@@ -0,0 +1,4 @@
+<!-- Licensed to the .NET Foundation under one or more agreements. The .NET Foundation licenses this file to you under the MIT license. See the License.txt file in the project root for more information. -->
+<Project>
+  <Import Project="..\build\$(MSBuildThisFileName).props"/>
```
tmat, line 3: >Import [](http://example.com/codeflow?start=3&length=6)

Why do we need buildTransitive? Seems like a project that depends on a project that has Source Link may or may not opt-into having Source Link. Either way it should work as there is no cross-project dependency on source link.

eng/BuildTask.targets
ViktorHofer: Fixed

src/Microsoft.Build.Tasks.Git/buildTransitive/Microsoft.Build.Tasks.Git.props
ViktorHofer: buildTransitive is necessary so that packages that reference Microsoft.Build.Tasks.Git.nupkg receive the msbuild props and targets files from it transitively. That's necessary when using NuGet's Pack task as it emits a slightly different nuspec as when hand-authoring it.

https://github.com/NuGet/Home/issues/6091 has all the details.

ViktorHofer:
> Would you mind listing the differences?

Sure. I will post the nuspec diffs next week.

ViktorHofer:
@tmat now I really finally got to finishing this PR.

I just diffed the package contents before and after this change and noticed a bug that resulted in assemblies being copied into the package layout unnecessarily. This is now fixed and the diff doesn't show any unexpected changes. The expected ones are:
- net472 folder name -> netframework
- core folder name -> net
- The nuspec now lists dependencies per TFM (that's the default behavior when using the NuGet pack task). Aside from a more verbose listing, this doesn't harm.
- The License.TXT file is now included (arcade setting)
- The build task assemblies now include their .deps.json / .config files. That's expected and the msbuild team recommends shipping those along an msbuild package.
- My local commit is different

### Example files diff:

image

(Ignore the package, _rels folder and the ContentType.xml file -> those are things that NuGet creates automatically)

### Example nuspec diff:

image

ViktorHofer:
@tmat when you are back from holiday, please take a look. Thanks in advance.

ViktorHofer:
@tmat gentle ping

eng/BuildTask.targets
```diff
@@ -0,0 +1,112 @@
+<!-- Copied from arcade with small modifications. -->
```
tmat, line 1: It'd be nice to share this rather than making a copy.

src/SourceLink.Tools/Microsoft.SourceLink.Tools.Package.csproj
```diff
@@ -1,19 +1,12 @@
-<Project Sdk="Microsoft.NET.Sdk">
```
tmat, line 1: We do want this project to compile the source. This is to validate that the source package code is correct and builds against the listed TFMs.

ViktorHofer: Makes sense. Reverted the change.

ViktorHofer:
@tmat PTAL

ViktorHofer:
gentle ping @tmat

tmat (approved):

PR #1144, opened 2023-11-16
Enable Source Link in design-time build

tmat:
Source Link, which is now included in the SDK and enabled by default, is causing Hot Reload to block stepping. Depending on timing of design-time builds this intermittently prevents the customer from debugging their application. Hot Reload blocks stepping due to differences in generated AssemlyInfo.cs files caused by Source Link targets.

Git commit SHA that is now included in `AssemblyInformationalVersion` attribute by default was not set in design-time build. This caused the content of `*.AssemlyInfo.cs` to differ between regular build and design-time build. If design-time build run after the debugging has started Hot Reload registered this difference as a change to be applied. Changing assembly level attributes is currently a rude edit and thus the change application is blocked, which in turns blocks stepping.

Fixes https://github.com/dotnet/sdk/issues/36666

## Customer Impact

Debugging projects is broken.

A workaround is to set `EnableSourceControlManagerQueries` to true globally (for all projects in the repo), e.g. add `Directory.Build.props` to the source root with:

```xml
<Project>
  <PropertyGroup>
    <EnableSourceControlManagerQueries>true</EnableSourceControlManagerQueries>
  </PropertyGroup>
</Project>
```

## Regression?

- [x] Yes
- [ ] No

7.0

## Risk

- [ ] High
- [ ] Medium
- [x] Low

The change sets property that enables the same code path that's already executed in regular build to run in design-time build.

## Verification

- [x] Manual (required)
- [x] Automated

## Packaging changes reviewed?

- [ ] Yes
- [x] No
- [ ] N/A

### DIFF
--- a/src/SourceLink.Common/build/Microsoft.SourceLink.Common.props
+++ b/src/SourceLink.Common/build/Microsoft.SourceLink.Common.props
@@ -7,14 +7,13 @@
 
   <PropertyGroup>
     <!--
-      Do not query source control manager for any information when building in the IDE or for Live Unit Testing.
-      Features that need this information will be disabled if false.
+      Used to suppress querying source control and features that use the information (e.g. git commit SHA).
     -->
-    <EnableSourceControlManagerQueries Condition="'$(EnableSourceControlManagerQueries)' == '' and '$(DesignTimeBuild)' != 'true' and '$(BuildingForLiveUnitTesting)' != 'true'">true</EnableSourceControlManagerQueries>
-    
+    <EnableSourceControlManagerQueries Condition="'$(EnableSourceControlManagerQueries)' == ''">true</EnableSourceControlManagerQueries>
+
     <!--
       Do not generate SourceLink when building in the IDE or for Live Unit Testing.
     -->
     <EnableSourceLink Condition="'$(EnableSourceLink)' == '' and '$(DesignTimeBuild)' != 'true' and '$(BuildingForLiveUnitTesting)' != 'true'">true</EnableSourceLink>
-  </PropertyGroup>  
+  </PropertyGroup>
 </Project>
--- a/src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
+++ b/src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets
@@ -14,11 +14,14 @@
     Triggers SetEmbeddedFilesFromSourceControlManagerUntrackedFiles target defined by a source control package Microsoft.Build.Tasks.{Git|Tfvc|...}.
     Assumes that all targets that generate source files and add them to Compile items run before BeforeCompile target.
     This is a convention established by common targets.
+    
+    Disabled for design-time build since calculating untracked files is non-trivial operation
+    and embedding them only affects the content of the generated PDB, which has no impact on design-time build.
   -->
   <Target Name="_SetEmbeddedFilesFromSourceControlManagerUntrackedFiles"
           DependsOnTargets="BeforeCompile;SetEmbeddedFilesFromSourceControlManagerUntrackedFiles"
           BeforeTargets="CoreCompile"
-          Condition="'$(EmbedUntrackedSources)' == 'true' and '$(EmbedAllSources)' != 'true' and '$(DebugType)' != 'none' and '$(EnableSourceControlManagerQueries)' == 'true'" />
+          Condition="'$(EmbedUntrackedSources)' == 'true' and '$(EmbedAllSources)' != 'true' and '$(DebugType)' != 'none' and '$(EnableSourceControlManagerQueries)' == 'true' and '$(DesignTimeBuild)' != 'true'" />
 
   <!--
     If defined populates MappedPath metadata of SourceRoot items.
--- a/src/SourceLink.Git.IntegrationTests/TargetTests.cs
+++ b/src/SourceLink.Git.IntegrationTests/TargetTests.cs
@@ -90,6 +90,7 @@ namespace Microsoft.SourceLink.IntegrationTests
                 expectedResults: new[]
                 {
                     NuGetPackageFolders,
+                    ProjectSourceRoot,
                     "true",
                     "",
                     ""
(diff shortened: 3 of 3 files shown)


src/SourceLink.Common/build/Microsoft.SourceLink.Common.props
```diff
@@ -7,14 +7,13 @@
...
+      Used to suppress querying source control and features that use the information (e.g. git commit SHA).
     -->
-    <EnableSourceControlManagerQueries Condition="'$(EnableSourceControlManagerQueries)' == '' and '$(DesignTimeBuild)' != 'true' and '$(BuildingForLiveUnitTesting)' != 'true'">true</EnableSourceControlManagerQueries>
-    
+    <EnableSourceControlManagerQueries Condition="'$(EnableSourceControlManagerQueries)' == ''">true</EnableSourceControlManagerQueries>
+
     <!--
       Do not generate SourceLink when building in the IDE or for Live Unit Testing.
     -->
     <EnableSourceLink Condition="'$(EnableSourceLink)' == '' and '$(DesignTimeBuild)' != 'true' and '$(BuildingForLiveUnitTesting)' != 'true'">true</EnableSourceLink>
```
tmat, line 17: @rainersigwald I kept this suppress in DTB since this only affects generated SourceLink.json that only impacts emit, so it should not be needed in DTB. Makes sense? Or should we enable it as well to have 100% consistency between DTB and command line build?

src/SourceLink.Common/build/Microsoft.SourceLink.Common.props
```diff
@@ -7,14 +7,13 @@
 
   <PropertyGroup>
     <!--
-      Do not query source control manager for any information when building in the IDE or for Live Unit Testing.
-      Features that need this information will be disabled if false.
+      Used to suppress querying source control and features that use the information (e.g. git commit SHA).
     -->
-    <EnableSourceControlManagerQueries Condition="'$(EnableSourceControlManagerQueries)' == '' and '$(DesignTimeBuild)' != 'true' and '$(BuildingForLiveUnitTesting)' != 'true'">true</EnableSourceControlManagerQueries>
```
rainersigwald, line 13: This seems expensive to run at design time, so I'd want someone from project-system to agree this is OK. @kvenkatrajan can you route?

rainersigwald, line 13: The LUT condition still seems reasonable--is it affected by the same set of problems? Is there signoff from the LUT team?

rainersigwald (commented):
The change looks fine to me but has implications for partner teams that I'd want signoff from.

src/SourceLink.Common/build/Microsoft.SourceLink.Common.props
rainersigwald: I think this makes sense to me.

src/SourceLink.Common/build/Microsoft.SourceLink.Common.props
tmat: If there is a test that reads AssemblyInformationalVersion it would behave differently when running LUT. I think it better if they are the same between DTB and command line build.

kvenkatrajan: @tmeschter please review.

tmat: To retrieve commit sha we do read a couple of files. However, we cache repository information in msbuild cache: https://github.com/dotnet/sourcelink/blob/4ca4236c10047416251be48339d464f15a697e8b/src/Microsoft.Build.Tasks.Git/RepositoryTask.cs#L96-L105

tmat: Actually, there is a potential for more IO. Embedding untracked sources is more expensive as it needs to iterate over all Compile items.

```
  <Target Name="_SetEmbeddedFilesFromSourceControlManagerUntrackedFiles"
          DependsOnTargets="BeforeCompile;SetEmbeddedFilesFromSourceControlManagerUntrackedFiles"
          BeforeTargets="CoreCompile"
          Condition="'$(EmbedUntrackedSources)' == 'true' and '$(EmbedAllSources)' != 'true' and '$(DebugType)' != 'none' and '$(EnableSourceControlManagerQueries)' == 'true'" />
```

Since embedding only affects PDB I think it would make sense to disable this target in DTB.

src/SourceLink.Common/build/Microsoft.SourceLink.Common.props
tmeschter: Is the SourceLink.json passed in as a compiler input? E.g. as an addtional item?

tmeschter (approved):
In general the set of compiler inputs and their contents should be the same in design-time builds and real builds. When we need to turn something off for DTBs we need to think hard about how that is going to impact the IDE. My advice is _don't_ disable things in the DTB until you know they are expensive.

src/SourceLink.Common/build/Microsoft.SourceLink.Common.props
tmat: It's a separate command line switch `/sourcelink`.

tmat:
> In general the set of compiler inputs and their contents should be the same in design-time builds and real builds. When we need to turn something off for DTBs we need to think hard about how that is going to impact the IDE. My advice is _don't_ disable things in the DTB until you know they are expensive.

Agreed, in general, but I think we can safely exclude command line switches that only affect emit and do not affect EnC.

tmeschter:
> Agreed, in general, but I think we can safely exclude command line switches that only affect emit and do not affect EnC.

Sounds good to me.

src/SourceLink.Common/build/Microsoft.SourceLink.Common.props
rainersigwald: The LUT team agreed over IM.

rainersigwald (approved):

tmat:
Fixed in 8.0.200-preview.23604.9

gitc
ViktorHofer, line 1: @tmat was that file accidentally checked in?

tmat: Oops, yes.

PR #1150, opened 2023-12-01
Fixes handling of local and network paths in remote URL

tmat:
Fixes https://developercommunity.visualstudio.com/t/VS-178:-Unable-to-locate-repository-wit/10519177 and https://developercommunity.visualstudio.com/t/Repository-URL-evaluation-exceeds-maximu/10523155

The implementation didn't correctly handle scenario where a repository is cloned from a _bare_ repository located on a local path or a network share. Source Link in this scenario can't be easily supported by the debugger and we should disable it without failing the build.

## Customer Impact

Build in repositories that are cloned from a file share fails after the customer upgrades their toolset to .NET 8 SDK.

## Regression?

- [x] Yes
- [ ] No

7.0

## Risk

- [ ] High
- [ ] Medium
- [x] Low

## Verification

- [x] Manual (required)
- [x] Automated

## Packaging changes reviewed?

- [ ] Yes
- [x] No
- [ ] N/A


tmat:
Fixed in 8.0.200-preview.23628.6.

PR #1153, opened 2023-12-07
Remove Microsoft.SourceBuild.Intermediate from prebuilt baseline

ellahathaway:
Prebuilt detection no longer detects Microsoft.SourceBuild.Intermediates as prebuilts due to https://github.com/dotnet/arcade/pull/13935. 

Addresses https://github.com/dotnet/source-build/issues/3010

### DIFF
--- a/eng/SourceBuildPrebuiltBaseline.xml
+++ b/eng/SourceBuildPrebuiltBaseline.xml
@@ -3,6 +3,5 @@
 
 <UsageData>
   <IgnorePatterns>
-    <UsagePattern IdentityGlob="Microsoft.SourceBuild.Intermediate.*" />
   </IgnorePatterns>
 </UsageData>


ellahathaway:
@tmat @MichaelSimons Could you take a look at the changes in this PR when you get a chance, thanks!

ellahathaway:
I also wanted to follow up on https://github.com/dotnet/sourcelink/pull/1038#issuecomment-1588713967 - if you are alright with granting write access then I will add the `source-build-internals` to the `CODEOWNERS` file.

MichaelSimons (approved):

tmat (approved):

tmat:
Can we delete the file entirely?

ellahathaway:
> Can we delete the file entirely?

After looking at https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.SourceBuild/tasks/src/UsageReport/ValidateUsageAgainstBaseline.cs#L24, it appears that this file can technically be deleted and prebuilt detection would default to comparing against `<UsageData />`. 

That said, IMO the effort involved in reintroducing the file later, should a prebuilt be detected and deemed acceptable, surpasses the advantages of deleting the file now.

PR #1248, opened 2024-07-11
Populate RepositoryBranch

### LINKED ISSUE #1243: Populate RepositoryBranch automatically
Following up from #188 and starting the discussion to populate `RepositoryBranch` automatically.

There are a few linked items in the old issue requesting this feature. Additionally, a new version of the DotNet.ReproducibleBuilds package was released, which sets this for some common CI providers (see https://github.com/dotnet/reproducible-builds/pull/27). However, that has the side-effect of causing the nuspec to be different between a local build and a CI build (beyond what's already necessary).

I'm open to design discussions, but I'll propose an algorithm that's the same as the one currently provided by ReproducibleBuilds, with a fallback that's essentially `git rev-parse --abbrev-ref HEAD` (or similar) if no properties are provided.

MattKotsenas:
Fixes #1243.

Adds `BranchName` as an output of the `LocateRepository` task and also adds it to the SourceRoot for the main repository. The locate task maps the `BranchName` output task to `SourceBranchName` in the .targets file to follow the naming convention. The branch is the full symbolic-ref name and not the short name to avoid ambiguity in source control systems like GitHub which use the `refs/pull` namespace for Pull Requests.

This change doesn't add branch name to submodules, as fundamentally submodules only track commits, and the main branch will fully specify any submodule-related metadata.

The corresponding change to map this value into `RepositoryBranch` as part of NuGet pack is @ https://github.com/NuGet/NuGet.Client/pull/5923.

Open issues in the PR:

1. [ ] Some of the `CloudHostedProvidersTests.cs` specifically use the inbox version of the Git tasks, and as such they can't assert the new functionality. I'm unsure if that's by design and I should not worry about that scenario, or if the tests should be changed?
2. [ ] Open to name changes anywhere; I went with the simple name, but happy to rename if there's something more appropriate.

### DIFF
--- a/src/Microsoft.Build.Tasks.Git/GitDataReader/GitReferenceResolver.cs
+++ b/src/Microsoft.Build.Tasks.Git/GitDataReader/GitReferenceResolver.cs
@@ -157,7 +157,7 @@ namespace Microsoft.Build.Tasks.Git
         {
             string reference = ReadReferenceFromFile(Path.Combine(_gitDirectory, GitRepository.GitHeadFileName));
 
-            return TryGetReferenceName(reference, out string? name) ? name : null;
+            return TryGetReferenceName(reference, out var name) ? name : null;
         }
 
         public string? ResolveReference(string reference)
@@ -185,7 +185,7 @@ namespace Microsoft.Build.Tasks.Git
         {
             // See https://git-scm.com/docs/gitrepository-layout#Documentation/gitrepository-layout.txt-HEAD
 
-            if (TryGetReferenceName(reference, out string? symRef))
+            if (TryGetReferenceName(reference, out var symRef))
             {
                 if (lazyVisitedReferences != null && !lazyVisitedReferences.Add(symRef))
                 {
--- a/src/Microsoft.Build.Tasks.Git/GitDataReader/GitReferenceResolver.cs
+++ b/src/Microsoft.Build.Tasks.Git/GitDataReader/GitReferenceResolver.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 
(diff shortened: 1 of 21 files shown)


MattKotsenas:
The test failure is because the integration tests are dependent on git's init.defaultBranch config. I'll look into making that config agnostic.

baronfel:
I haven't looked in detail, but to answer the question about where the package metadata mapping should go I'd say a followup PR to NuGet would be the appropriate place - they already have the mappings for the other automatic package properties.

MattKotsenas:
> I haven't looked in detail, but to answer the question about where the package metadata mapping should go I'd say a followup PR to NuGet would be the appropriate place - they already have the mappings for the other automatic package properties.

OK, I can get started on that. Separately but relatedly, do you know the flow for updates? I'm new in this area, but to "ship" this feature end-to-end it sounds like we'd need at least:

1. Update SourceLink package
2. Update NuGet client package
3. SDK update to pull in new versions of 1 + 2

Is this a manual process, or are these picked up automatically? Any info greatly appreciated. Thanks!

baronfel:
There's automated code flow of both SourceLink and NuGet targets into the SDK now - if you're working against main branch for all of the repos then you're targeting .NET 9 at this point, which feels fine to me.

MattKotsenas:
Created a draft in NuGet.Client: https://github.com/NuGet/NuGet.Client/pull/5923.

@tmat, would you mind taking a look, or letting me know who can review? Happy to make any changes requested. Thanks!

tmat:
I'll take a look.

src/Microsoft.Build.Tasks.Git/GitDataReader/GitReferenceResolver.cs
```diff
@@ -152,23 +153,40 @@ public GitReferenceResolver(string gitDirectory, string commonDirectory)
         public string? ResolveHeadReference()
             => ResolveReference(ReadReferenceFromFile(Path.Combine(_gitDirectory, GitRepository.GitHeadFileName)));
 
+        public string? GetBranchForHead()
+        {
+            string reference = ReadReferenceFromFile(Path.Combine(_gitDirectory, GitRepository.GitHeadFileName));
+
+            return TryGetReferenceName(reference, out string? name) ? name : null;
```
tmat, line 160: Nit: you can use `out var` here

src/SourceLink.Git.IntegrationTests/CloudHostedProvidersTests.cs
```diff
@@ -76,12 +76,14 @@ public void NoRepository_NoWarnings()
                 expressions: new[]
                 {
                     "@(SourceRoot)",
+                    // "@(SourceRoot->'%(BranchName)')", // TODO: Unclear what to do here since this test explicitly uses the SDK version of Microsoft.Build.Tasks.Git
```
tmat, line 79: Let's file an issue and link it from the TODO comment. Once the SDK is updated with new Source Link version this should work, right?

tmat:
@MattKotsenas Looks great! Thanks for the PR. Just minor update to the comments and it's ready to go.

src/SourceLink.Git.IntegrationTests/CloudHostedProvidersTests.cs
MattKotsenas: Done!

src/Microsoft.Build.Tasks.Git/GitDataReader/GitReferenceResolver.cs
MattKotsenas: Fixed!

MattKotsenas:
Thanks @tmat! I believe I fixed all your comments. Would you mind taking another look?

PR #1264, opened 2024-08-20
Use Central Package Management

ViktorHofer:
... and enable transitive pinning.

Define package dependencies in Directory.Packages.props. Set IsTestSupportProject in TestUtilities.csproj.


ViktorHofer:
@tmat when you are back from holiday, please take a look. Thanks in advance.

ViktorHofer:
@tmat gentle ping

tmat (approved):
Thanks!

PR #1271, opened 2024-08-28
Refactor "PDB distributions" section of README

### LINKED ISSUE #255: guidance for embedding pdb in nupkg vs snupkg
I'd like to revisit the change in guidance from https://github.com/dotnet/sourcelink/commit/f92b277cf08caf4b86425aad1fc4cdebd63c6650 as a fix to #229.

> Including PDBs in the .nupkg is generally no longer recommended as it increases the size of the package and thus restore time for projects that consume your package, regardless of whether the user needs to debug through the source code of your library or not

This is only when publishing to NuGet.org. The only time a snupkg would be useful is when publishing to NuGet.org. If you are using any other NuGet feed such as:
- Azure Artifacts NuGet feed
- Artifactory NuGet feed
- MyGet.org feed
- AppVeyor feed
- many more...

Including PDBs in the .nupkg works all of those places and NuGet.org as well. It works at every step of development. It decreases the size of the nupkg download only slightly. The same can be said for embedded XML documentation.

## Symbol Servers need to retire
One of the most time frustrations for me is when Visual Studio debugger HTTP queries symbol servers for symbols when I want to debug something. I should be able to lock dependencies and have `dotnet restore` bring all of the available symbols asynchronously for all of the locked dependencies directly after a restore. It would be nice if all NuGet servers supported snupkg too and `dotnet restore` of symbols would look for snupkg first before querying Symbol Servers.

MattKotsenas:
Refactor the "PDB distributions" section of the README so that:

- Add a small pros / cons table explaining the benefits and limitations of: .snupkgs, Adding to the main package, and assembly embedding
- There's less bouncing back and forth between text (specifically shrinking the distance between introducing a .snupkg and talking about the limitations)
- Add explanation of embedding and benefits / limitations
- Update "main package" alternative to make it clear that it requires consumers to set `CopyDebugSymbolFilesFromPackages`
- Use GitHub markdown tags for asides or secondary information

Fixes #255


MattKotsenas:
I originally wrote this change as part of #1270, but separated it into its own PR since this one is a bit more nuanced and contains some opinions instead of just facts.

MattKotsenas:
@tmat, would you mind taking a look at this when you have a second? If you have any questions or concerns, please let me know. Thanks!

README.md
```diff
@@ -208,25 +208,36 @@ The VC++ linker supports `/SOURCELINK` [switch](https://docs.microsoft.com/en-us
...
+
+```xml
+<PropertyGroup>
+  <DebugType>embedded</DebugType>
+</PropertyGroup>
+```
+
+Alternatively, you can include PDB files in the main NuGet package by setting the following property in your project:
+
+```xml
```
tmat, line 233: I would suggest including portable PDBs in the nupkg first and embedding PDBs into the assembly as an alternative. The latter increases size of the assembly, which may potentially be an issue in some cases.

MattKotsenas: Can you expand on what the issues are? I can imagine that it slows down assembly loading. Anything else? I'll add it into the document as a caveat. Thanks!

MattKotsenas: Did an update assuming the assembly load costs are the main issue. As part of this update, I also added a small pros / cons table to make it easier to understand the main tradeoffs, then gave each alternative its own section.

README.md
```diff
@@ -226,27 +226,60 @@ The VC++ linker supports `/SOURCELINK` [switch](https://docs.microsoft.com/en-us
...
 
+There are three main ways to distribute PDBs: via a dedicated .snupkg, including them in the main package, and embedding them directly into the assembly. Each has advantages and drawbacks, depending on your use case.
+
+|                                         | snupkg                      | Include in main package | Embed in assembly |
+|-----------------------------------------|-----------------------------|-------------------------|-------------------|
+| No user opt-in required to load symbols | ✅                          | ❌                      | ✅                |
+| No increase in size of main package     | ✅                          | ❌                      | ❌                |
+| No increase in size of assemblies       | ✅                          | ✅                      | ❌                |
+| Supported by all package feeds          | ❌ (Supported on NuGet.org) | ✅                      | ✅                |
+| Supports all symbol types               | ❌ (Portable symbols only)  | ✅                      | ✅                |
```
tmat, line 237: Only Portable PDB can be embedded in assembly. 

Not sure I would even include this line - Windows PDB are legacy for managed code. They should only be used for C++.

README.md
```diff
@@ -226,27 +226,60 @@ The VC++ linker supports `/SOURCELINK` [switch](https://docs.microsoft.com/en-us
...
 </PropertyGroup>
 ```
 
-Keep in mind that including PDBs in the .nupkg increases the size of the package and thus restore time for projects that consume your package, regardless of whether the user needs to debug through the source code of your library or not.
+> [!IMPORTANT]
+> Keep in mind that including debug information in the .nupkg increases the size of the package and thus restore time for projects that consume your package, regardless of whether the user needs to debug the source code of your library or not.
 
-.snupkg symbol packages have following limitations:
+> [!IMPORTANT]
+> When including PDB files in the main package, projects that consume the package must _also_ opt-in to copying the symbols into their own output directory. Starting in .NET 7 this can be controlled via the [`CopyDebugSymbolFilesFromPackages`](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#copydebugsymbolfilesfrompackages) property.
```
tmat, line 266: @baronfel Shouldn't this property be true by default?

README.md
```diff
@@ -226,27 +226,60 @@ The VC++ linker supports `/SOURCELINK` [switch](https://docs.microsoft.com/en-us
...
+```xml
+<PropertyGroup>
+  <DebugType>embedded</DebugType>
+</PropertyGroup>
+```
 
-Consider including PDBs in the main package if it is not possible to use .snupkg for the above reasons. 
-For managed projects, consider switching to Portable PDBs by setting `DebugType` property to `portable`. This is the default for .NET SDK projects, but not classic .NET projects.
+> [!IMPORTANT]
+> Keep in mind that embedding debug including debug information in assembly both increases the size of the package and thus restore time for projects that consume your package, regardless of whether the user needs to debug the source code of your library or not, and also may impact assembly load performance for very large assemblies.
```
tmat, line 283: ... and size of applications that deploy the binaries.

README.md
baronfel: https://github.com/dotnet/sdk/pull/27580 - it was added in this PR and no default was set (so effectively `false`) because it would bloat application sizes. We could potentially make another push here, but we didn't feel that there was enough justification at the time.

README.md
MattKotsenas: Removed the line and added a word to note that this section is for managed code.

MattKotsenas: I believe I've addressed your concerns, so I'm marking this thread as resolved. If you disagree or have any other issues, please reactivate and provide updated details. Thanks!

README.md
MattKotsenas: I believe I've addressed your concerns, so I'm marking this thread as Resolved. If you disagree or have additional questions, please reactivate and let me know. Thanks!

README.md
MattKotsenas: Added!

Reworded this section a bit to make it clearer as well.

MattKotsenas: I believe I've addressed your concerns, so marking this thread as Resolved. If you disagree or have any questions, please reactivate and let me know. Thanks!

baronfel (approved):

README.md
tmat: @baronfel Not sure how does it bloat app size. Does publishing automatically include PDBs that are in the bin directory? I would hope not.

baronfel: Yes, all of the `ReferenceCopyLocalPaths` are copied to publish output by default: https://github.com/dotnet/sdk/blob/2c86f58b098147e268e4a19ce52294128f0b4b2a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets#L683-L695

tmat: Is that expected? I'd not expect PDBs getting published in release build.
If someone sets `CopyDebugSymbolFilesFromPackages` to `true`, how do they avoid publishing PDBs in the app?

baronfel: The same way you prevent anything being copied - setting the `Private="true"` metadata on the `Content` items. But yes, this is a fun area overall, and the interaction of these Items is non-intuitive.

MattKotsenas: @tmat, anything else here? Is this OK to merge?

tmat (approved):

PR #1273, opened 2024-08-29
Add CODEOWNERS

MattKotsenas:
People without write access to the repo (like me) can't add reviewers directly. Instead, we must tag people for review, which is easy to miss under the volume of notifications.

I've added a `CODEOWNERS` file with PR approvers from the last 6 months (@tmat and @ViktorHofer).

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,2 @@
+# These owners will be the default owners for everything unless a later match takes precedence
+* @tmat @ViktorHofer


MattKotsenas:
@tmat / @ViktorHofer, I picked you two based on the last 6 months of repo history. If there's better choices or a group to use please let me know. Thanks!

tmat (approved):

PR #1486, opened 2025-10-19
Update Microsoft.TeamFoundationServer.ExtendedClient

tmat:


eng/Versions.props
```diff
@@ -8,7 +8,7 @@
     <!-- Opt-in repo features -->
     <UsingToolSymbolUploader>true</UsingToolSymbolUploader>
     <!-- Prohibit the usage of .NET Standard 1.x dependencies. -->
-    <FlagNetStandard1XDependencies>true</FlagNetStandard1XDependencies>
```
tmat: Had to disable this to get the build working:

```
Microsoft.Build.Tasks.Tfvc net472 failed with 1 error(s) (0.1s)
The following .NET Standard 1.x packages are referenced and must be removed:
      - System.Text.Encoding
```

Issue #1450, opened 2025-06-25
Add RevisionTimeStamp as an MSBuild property


zivkan:
It could be useful to know what a commit's timestamp was, in addition to the commit hash/id.

This will allow downstream tooling to use it as appropriate. For example, some basic deployment tools only check file timestamps, so it could be interesting to modify files last write time to the commit timestamp, in order to get deterministic outputs, but still compatible with tools that only use timestamps.

I suggest the timestamp should be in ISO8601 format, possibly with an option to force a specific timezone (commit's original timezone or current machine's local timezone, or UTC)

PR #1552, opened 2025-12-01
Add RevisionTimestamp

omajid:
Add a new MSBuild property called RevisionTimestamp that contains a timestamp indicating when the latest commit was added to git. This can then be used to timestamp the produced binaries (eg, dll files, file creation times of the contents of nuget packages) to make .NET builds and packages more deterministic/reproducible.

The timestamp is in the ISO 8601/RFC 3339 format, as generated by the "O" formatter.

Fixes: #1450


omajid:
cc @zivkan

Issue #1290, opened 2024-10-02
"Repository has no commit" error in case of worktree + packed-refs


vikukush:
After git decided to pack my repo with several worktrees and a bunch of local branches, my builds started failing with following error:

D:\Corextroot\microsoft.build.tasks.git\8.0.0\build\Microsoft.Build.Tasks.Git.targets(25,5): Repository has no commit. [D:\Pl3\Test\UnitTests\Common\Tools.UnitTests.Common\Tools.UT.Common.csproj]

- D:\Pl3 is a worktree of D:\Platform.
- D:\Platform\.git\worktrees\Pl3\refs is empty.
- D:\Platform\.git\refs\heads doesn't have a ref of a branch that's checked out in D:\Pl3 either
- The only reference to the commit ID of a branch specified in D:\Platform\.git\worktrees\Pl3\HEAD is in D:\Platform\.git\packed-refs

Sourcelink task is unable to handle this situation.

I have worked to add a unit test that covers my scenario: https://github.com/dotnet/sourcelink/pull/1289

vikukush:
The issue went away after I created the very next commit, which created D:\Platform\.git\refs\heads\user\vikukush\mstestv3 file, so the commit ID was no longer inside packed-refs.
So the bug can be easily worked around, but this of course still needs to be fixed.

bradwilson:
I just started using `git worktree` and ran into this.

Any progress here, or even an acknowledgment from the team that it's an issue to be fixed?

tmat:
Will take a look.

tmat:
The path `D:\Corextroot\microsoft.build.tasks.git\8.0.0` indicates that you're using .NET SDK 8.0. Does the issue reproduce with .NET SDK 10?

jjonescz:
I'm hitting this too. I'm using Microsoft.Build.Tasks.Git from .NET SDK 10.0.105. It seems this happens when the HEAD ref is packed, i.e., there is no file under refs/heads/ for it in .git folder, it's only in .git/packed-refs. Can be worked around by creating a new commit (since that one won't be packed yet).

tmat:
@jjonescz Do you have repro steps how do you get the repo to that state?

jjonescz:
[...]

PR #1289, opened 2024-10-02
Add Worktree_Packed_GetHeadCommitSha unit test

vikukush:
Add test covering situation where worktree refs are packed.
In this case none of the refs/ folders contain files with commit ids, they exist only in packed-refs file in main .git folder.
See #1290

PR #1657, opened 2026-04-21
Use common dir to locate packed-refs and reftable

tmat:
The docs https://git-scm.com/docs/git-pack-refs say the `packed-refs` file is in `$GIT_DIR`, but it's actually in `$GIT_COMMON_DIR`. This is mentioned in https://git-scm.com/docs/gitrepository-layout#Documentation/gitrepository-layout.txt-packed-refs.

Similarly, `reftable` is also located in `$GIT_COMMON_DIR` even though [the doc](https://git-scm.com/docs/reftable#_layout) says it's in `$GIT_DIR`.

Fixes https://github.com/dotnet/sourcelink/issues/1290
Includes https://github.com/dotnet/sourcelink/pull/1289 by @vikukush


tmat:
@jjonescz ptal

jjonescz (approved):

PR #1722, opened 2026-06-16
Update msbuild to 18.6.3

tmat:
Note: https://github.com/dotnet/sdk/blob/main/src/Layout/redist/minimumMSBuildVersion should be updated to 18.6.3

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -15,12 +15,12 @@
 
   <PropertyGroup>
     <!-- msbuild -->
-    <MicrosoftBuildVersion>17.11.48</MicrosoftBuildVersion>
-    <MicrosoftBuildTasksCoreVersion>17.11.48</MicrosoftBuildTasksCoreVersion>
+    <MicrosoftBuildVersion>18.6.3</MicrosoftBuildVersion>
+    <MicrosoftBuildTasksCoreVersion>18.6.3</MicrosoftBuildTasksCoreVersion>
     <!-- nuget -->
     <NuGetVersioningVersion>6.12.1</NuGetVersioningVersion>
     <!-- runtime -->
-    <SystemTextJsonVersion>8.0.5</SystemTextJsonVersion>
+    <SystemTextJsonVersion>10.0.3</SystemTextJsonVersion>
   </PropertyGroup>
 
 </Project>
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -20,7 +20,7 @@
     <!-- nuget -->
     <NuGetVersioningVersion>6.12.1</NuGetVersioningVersion>
     <!-- runtime -->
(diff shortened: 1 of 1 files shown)


eng/Versions.props
```diff
@@ -15,12 +15,12 @@
...
     <!-- msbuild -->
-    <MicrosoftBuildVersion>17.11.48</MicrosoftBuildVersion>
-    <MicrosoftBuildTasksCoreVersion>17.11.48</MicrosoftBuildTasksCoreVersion>
+    <MicrosoftBuildVersion>18.4.0</MicrosoftBuildVersion>
+    <MicrosoftBuildTasksCoreVersion>18.4.0</MicrosoftBuildTasksCoreVersion>
     <!-- nuget -->
     <NuGetVersioningVersion>6.12.1</NuGetVersioningVersion>
     <!-- runtime -->
-    <SystemTextJsonVersion>8.0.5</SystemTextJsonVersion>
+    <SystemTextJsonVersion>10.0.3</SystemTextJsonVersion>
```
ViktorHofer: I would pick latest here


```suggestion
    <SystemTextJsonVersion>10.0.9</SystemTextJsonVersion>
```

ViktorHofer (approved):

ViktorHofer:
This will introduce a prebuilt in the VMR as that version isn't available in source-build-assets. @dotnet/source-build does it make sense to use PSB here (define the MSBuild dependencies in Version.Details.xml) or should we add the new version to source-build-assets? Those are compile time only dependencies which don't get redistributed.

ViktorHofer:
Still interested in guidance here regarding whether to prefer SBRP or PSB in such a case, cc @mthalman @MichaelSimons
I see that arcade i.e. defines them so that they are coming from PSB: https://github.com/dotnet/arcade/blob/c99e4a3ff820236cdcf303a084556adf54a4e8f3/eng/Version.Details.xml#L131-L147

mthalman:
This was already addressed in https://github.com/dotnet/dotnet/pull/7306/changes/e14aa12b3a6ec46c1a3a0d5228ca74810232c576#diff-b38df83bd311458adb67f2ed9accf1ebb7676283d88c9c427777d597506f90bf

ViktorHofer:
I was interested in when to use PSB vs SBRP - in particular for this case. Arcade uses PSB for the msbuild dependencies, not SBRP.

mthalman:
PSB would be used when the assembly needs to be executed during the build, as opposed to just compiled against for SBRP.

ViktorHofer:
With SBRP, there's still the manual work of adding the packages to the repository. Is there a strong disadvantage in using the PSB artifact instead that makes us not choose that option?

mthalman:
In order for anything to exist in PSB, it needs to have been produced as output of the previous release. So that's the limitation.

PR #1732, opened 2026-06-30
Migrate AzureRepos.Git.GetSourceLinkUrl to multithreaded task model

jankratochvilcz:
## Summary
- Adds `[MSBuildMultiThreadableTask]` to `Microsoft.SourceLink.AzureRepos.Git.GetSourceLinkUrl`.
- Implements `IMultiThreadableTask` with `TaskEnvironment.Fallback` so direct unit-test construction preserves current behavior.
- Replaces the `BUILD_REPOSITORY_URL_MAP*` lookup in `TryGetEnvironmentUriMap()` with `TaskEnvironment.GetEnvironmentVariable(...)` for MT-safe environment access.

## Audit
- `grep -rn "TryGetEnvironmentUriMap" src/` only finds the private method declaration in `src/SourceLink.AzureRepos.Git/GetSourceLinkUrl.cs`; it is currently uncalled dead code with an existing TODO. Because there is no active behavior to exercise, no new behavioral unit test was added.
- Verified no direct `Environment.*`, `Directory.GetCurrentDirectory`, or `Path.GetFullPath` usage remains in this file.
- The environment-variable swap is behavior-preserving for existing single-process/direct construction because `TaskEnvironment.Fallback` reads the process environment.
- No output property or log message consumes an absolutized path in this change, so the path compatibility concerns from the MT migration checklist do not apply.

References dotnet/msbuild#14093.

## Validation
- `./build.sh --restore --build`
- `PATH="$PWD/.dotnet:$PATH" dotnet test src/SourceLink.AzureRepos.Git.UnitTests/Microsoft.SourceLink.AzureRepos.Git.UnitTests.csproj -f net11.0 --no-restore`

### DIFF
--- a/src/SourceLink.AzureRepos.Git/GetSourceLinkUrl.cs
+++ b/src/SourceLink.AzureRepos.Git/GetSourceLinkUrl.cs
@@ -11,9 +11,12 @@ using Microsoft.Build.Tasks.SourceControl;
 
 namespace Microsoft.SourceLink.AzureRepos.Git
 {
-    public sealed class GetSourceLinkUrl : GetSourceLinkUrlGitTask
+    [MSBuildMultiThreadableTask]
+    public sealed class GetSourceLinkUrl : GetSourceLinkUrlGitTask, IMultiThreadableTask
     {
         private const string UrlMapEnvironmentVariableName = "BUILD_REPOSITORY_URL_MAP";
+
+        public TaskEnvironment TaskEnvironment { get; set; } = TaskEnvironment.Fallback;
 
         protected override string HostsItemGroupName => "SourceLinkAzureReposGitHost";
         protected override string ProviderDisplayName => "AzureRepos.Git";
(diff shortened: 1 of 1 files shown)


jankratochvilcz:
Automated MT review (MT-migration reviewer + expert code review): both **approved**, no blocking issues. The only nit raised — dropping the `value!` null-forgiving operator as redundant — was investigated and is in fact **required**: `TaskEnvironment.GetEnvironmentVariable` is annotated `string?`, and net472's `string.IsNullOrEmpty` lacks `[NotNullWhen(false)]`, so the value isn't narrowed there (CS8604 without it). Kept it and added a clarifying comment.

JanProvaznik (approved):

tmat (changes requested):
TryGetEnvironmentUriMap is unused. You can delete it.

jankratochvilcz:
Thanks, done @tmat !

tmat (approved):

PR #1733, opened 2026-06-30
Migrate Common.GenerateSourceLinkFile to multithreaded task model

jankratochvilcz:
## Summary

Migrates `Microsoft.SourceLink.Common.GenerateSourceLinkFile` to the MSBuild multithreaded task model for dotnet/msbuild#14093.

## Details

- Adds `[MSBuildMultiThreadableTask]` and implements `IMultiThreadableTask` with `TaskEnvironment.Fallback` so direct unit-test construction keeps existing semantics.
- Resolves `OutputFile` via `TaskEnvironment.GetAbsolutePath` before all `File.Exists`, `File.Delete`, `File.ReadAllText`, and `File.WriteAllText` calls, so relative paths are based on the project directory instead of process CWD.
- Preserves Sin 1 behavior by keeping the `[Output] SourceLink` value as the original `OutputFile` string, not the absolute path.
- Preserves Sin 2 behavior by continuing to log the original `OutputFile` value and sanitizing exception messages that include the absolutized path back to the original path.
- Keeps the prior empty-content/null-or-empty `OutputFile` no-existing-file behavior before absolutization; non-empty content still fails through the existing catch path.

## Tests

- `./build.sh --restore --build --verbosity minimal`
- `./.dotnet/dotnet test src/SourceLink.Common.UnitTests/Microsoft.SourceLink.Common.UnitTests.csproj -f net11.0 --no-restore`
- Added `WriteSourceLinkFileResolvesRelativeOutputFileAgainstProjectDirectory`, a decoy-CWD test that verifies a relative output file is written under the `TaskEnvironment` project directory, not process CWD, while `SourceLink` remains the original relative value.

### DIFF
--- a/src/SourceLink.Common/GenerateSourceLinkFile.cs
+++ b/src/SourceLink.Common/GenerateSourceLinkFile.cs
@@ -13,8 +13,11 @@ using Microsoft.Build.Utilities;
 
 namespace Microsoft.SourceLink.Common
 {
-    public sealed class GenerateSourceLinkFile : Task
+    [MSBuildMultiThreadableTask]
+    public sealed class GenerateSourceLinkFile : Task, IMultiThreadableTask
     {
+        public TaskEnvironment TaskEnvironment { get; set; } = TaskEnvironment.Fallback;
+
         [Required, NotNull]
         public ITaskItem[]? SourceRoots { get; set; }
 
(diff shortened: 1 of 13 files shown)


jankratochvilcz:
Automated MT review (MT-migration reviewer + expert code review): both **approved**, no blocking/actionable findings. Verified: Sin 1 (`SourceLink` output keeps the original relative path on both the up-to-date and write branches), Sin 2 (all log/error messages use the original `OutputFile`; catch-block sanitizes the exception message back to the original path), Sin 3/6 (full `{content}×{OutputFile}` truth table matches pre-migration behavior; the added `content==null && IsNullOrEmpty(OutputFile)` guard avoids `GetAbsolutePath("")` while preserving the original outcome), nullable `AbsolutePath?` handling, and a parallel-safe decoy-CWD test that fails if the migration is reverted. No changes required.

JanProvaznik (approved):

OvesN (approved):

src/SourceLink.Common/GenerateSourceLinkFile.cs
```diff
@@ -112,19 +115,28 @@ private void WriteSourceLinkFile(string? content)
                 Log.LogWarning(Resources.SourceControlInformationIsNotAvailableGeneratedSourceLinkEmpty);
             }
 
+            if (content == null && string.IsNullOrEmpty(OutputFile))
```
tmat: I'd make it a requirement on `OutputFile` to be an absolute path. The value of this parameter is controlled by our target and set from `_SourceLinkFilePath` defined here: https://github.com/dotnet/sourcelink/blob/2f9f85141fe5ed8bb1b67ad0625636a2d43c280f/src/SourceLink.Common/build/Microsoft.SourceLink.Common.targets#L9

tmat (changes requested):

src/SourceLink.Common/GenerateSourceLinkFile.cs
jankratochvilcz: @tmat I changed the implementation, please check it out

tmat (approved):

src/SourceLink.Common/GenerateSourceLinkFile.cs
jankratochvilcz: @tmat Are you confident on this one? We have failing tests that suggest that currently absolute paths are not always being passed 👀

image

tmat (approved):

PR #1734, opened 2026-06-30
Migrate Tasks.Git.LocateRepository to multithreaded task model

jankratochvilcz:
## Summary

Migrates `Microsoft.Build.Tasks.Git.LocateRepository` (and its shared abstract base `RepositoryTask`) to the MSBuild multithreaded (MT) task model. The MT hazard: repository discovery resolved the initial path through `GitRepository.TryFindRepository`, which calls `Path.GetFullPath` internally — making it dependent on the process current working directory (CWD), which is shared across projects in the MT model.

## Approach: reject non-fully-qualified paths (attribute-only)

Rather than absolutizing the initial path via `TaskEnvironment.GetAbsolutePath`, this migration **rejects** any initial path that is not fully qualified. Repository discovery must not depend on the shared process CWD/drive; a relative, drive-relative (`C:foo`), or root-relative (`\foo`) path is rejected with the standard "missing repository" warning instead of being silently resolved against process state.

This keeps the migration **attribute-only**: the concrete `LocateRepository` carries `[MSBuildMultiThreadableTask]`, but the base `RepositoryTask` needs no `IMultiThreadableTask` / `TaskEnvironment` because it no longer performs any path absolutization — it only validates.

### `IsPathFullyQualified` polyfill
`Path.IsPathFullyQualified` does not exist on `net472`, so `PathUtilities.IsPathFullyQualified` provides a faithful reimplementation of the runtime's `PathInternal.IsPartiallyQualified` (Windows) under `#if NETFRAMEWORK`, delegating to the BCL API otherwise. It correctly rejects the CWD/drive-dependent forms that `Path.IsPathRooted` would accept (Sin 7). Covered by `PathUtilitiesTests` with `WindowsOnly`/`UnixOnly`-gated cases (drive-qualified, UNC, device `\\?\`, drive-relative, root-relative, invalid drive) plus OS-agnostic relative cases.

### Base `RepositoryTask`
`GetOrCreateRepositoryInstance()` guards the initial path: `if (string.IsNullOrEmpty(initialPath) || !PathUtilities.IsPathFullyQualified(initialPath)) { ReportMissingRepositoryWarning(initialPath); return null; }`. The `IsNullOrEmpty` short-circuit preserves the pre-migration graceful behavior for empty input and avoids the polyfill's `ArgumentNullException` on null. Warnings keep the **original** input for diagnostics (Sin 2).

### Concrete `LocateRepository`
Annotated with `[MSBuildMultiThreadableTask]` (`Inherited=false`, so each concrete class needs it). No `TaskEnvironment` API is required.

## Behavior change (intentional)
A relative initial path that previously resolved against the process CWD is now rejected → "missing repository" warning. In the shipped targets `LocateRepository` receives `$(MSBuildProjectDirectory)` (always fully qualified), so no production scenario is affected; the change only hardens against CWD-dependent inputs.

## Scope notes (from review)
- The guard delivers **current-directory / current-drive independence**. Machine/user-global git-config discovery in `GitEnvironment.CreateFromProcessEnvironment` still reads process env vars (`HOME`/`XDG_CONFIG_HOME`/`PROGRAMDATA`/`PATH`) — this is unchanged, read-only, and identical across all projects in a build, so it does not carry the per-project shared-state hazard the migration targets.
- The guard lives in the shared base `RepositoryTask`, so it also affects the sibling `GetUntrackedFiles` — which is migrated with the same approach in dotnet/sourcelink#1735 (and adds its own guard for its second `ProjectDirectory` consumption site).

## Merge coordination
`PathUtilities.cs`, `PathUtilitiesTests.cs`, `RepositoryTask.cs`, and the test-assembly `AssemblyInfo.cs` are **byte-identical** to the sibling PR dotnet/sourcelink#1735, so git auto-resolves "both branches added the same content" when the second PR merges — no conflict.

## Build & test
- `./build.sh --build` → 0 warnings, 0 errors (both `net472` and modern TFM compile).
- `dotnet test Microsoft.Build.Tasks.Git.UnitTests -f net11.0` → 448 passed, 4 skipped (Windows-only).
- `RelativePath_IsRejected_IndependentOfProcessCwd` fails if the migration is reverted (verified); `AbsolutePath_LocatesRepository` confirms historical behavior.

Related: dotnet/msbuild#14093.


jankratochvilcz:
## Review feedback addressed

Thanks for the thorough MT + expert reviews. Pushed `b0b4659`:

**MAJOR — empty/null `Path` regression (graceful → crash).** Confirmed and fixed. Post-migration, `TaskEnvironment.GetAbsolutePath("")` throws `ArgumentException` *before* `TryFindRepository`, and `ExecuteImpl`'s catch only handles `IOException/InvalidDataException/NotSupportedException`, so it propagated as MSB4018 — whereas pre-migration the same empty input hit `TryFindRepository`'s internal `Path.GetFullPath("")`, which threw the same exception but had it swallowed, yielding a graceful missing-repository warning and `Execute() == true`. Fix wraps the absolutize+discover step in `try { … } catch (ArgumentException) { ReportMissingRepositoryWarning(initialPath); return null; }` (`ArgumentNullException` derives from `ArgumentException`, so null is covered too).

**Empirical note on "whitespace":** I verified against Microsoft.Build 18.6.3 that `GetAbsolutePath` throws only on `null` and `""` — whitespace-only (`"   "`) does *not* throw (it's treated as a relative segment). So the reachable degenerate case is empty string, reachable via direct `Execute()` (the engine's `[Required]` validation doesn't run there).

**Inaccurate comment.** Rewritten to describe the *control-flow* equivalence (swallowed → warning → success), not just the matching exception type.

**Regression test.** Added `EmptyPath_DegradesGracefully_DoesNotThrow`. Dual-fault verified: it throws (fails) when the new catch is reverted, passes with the fix in place.

**MINOR — assembly-wide parallelization disable.** Kept (acknowledged acceptable); the CWD-mutating test makes process-global `Directory.SetCurrentDirectory` calls that must not race. Comment made test-agnostic.

All 437 `Microsoft.Build.Tasks.Git.UnitTests` (net11.0) pass; full build clean. Shared `RepositoryTask.cs`/`AssemblyInfo.cs` kept byte-identical with #1735.

src/Microsoft.Build.Tasks.Git/RepositoryTask.cs
```diff
@@ -112,9 +119,31 @@ private void ExecuteImpl()
                 return null;
             }
 
+            // Keep the original (possibly relative) path for user-facing messages (Sin 2).
             var initialPath = GetInitialPath();
 
-            if (!GitRepository.TryFindRepository(initialPath, out var location))
+            // Resolve the initial path against the task's project directory rather than the process
```
tmat: This is not necessary. The initial path is already an absolute path.
https://github.com/dotnet/sourcelink/blob/main/src/Microsoft.Build.Tasks.Git/build/Microsoft.Build.Tasks.Git.targets#L26

https://github.com/dotnet/sourcelink/blob/main/src/Microsoft.Build.Tasks.Git/build/Microsoft.Build.Tasks.Git.targets#L53

You can add a check that reports an error if it's not an absolute path.

tmat (changes requested):

src/Microsoft.Build.Tasks.Git.UnitTests/LocateRepositoryTests.cs
```diff
@@ -0,0 +1,107 @@
...
+using Microsoft.Build.Framework;
+using TestUtilities;
+using Xunit;
+
+namespace Microsoft.Build.Tasks.Git.UnitTests
+{
+    public class LocateRepositoryTests
+    {
+        /// <summary>
+        /// Verifies that <see cref="LocateRepository"/> resolves a relative <see cref="LocateRepository.Path"/>
```
tmat: Unnecessary.

src/Microsoft.Build.Tasks.Git/RepositoryTask.cs
jankratochvilcz: Thanks, I changed it as asked

tmat (approved):

tmat (changes requested):
Why do we need MockEngine4? There is already MockEngine defined in TestUtilities.

src/TestUtilities/MockEngine4.cs
```diff
@@ -0,0 +1,104 @@
...
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Build.Framework;
+
+namespace TestUtilities
+{
+    /// <summary>
+    /// An <see cref="IBuildEngine4"/> variant of <see cref="MockEngine"/>. Required by tasks that call the
```
tmat: The existing MockEngine should implement IBuildEngine4 if needed

jankratochvilcz: Done in 3b2d360. Removed `MockEngine4` and extended the existing `MockEngine` to implement `IBuildEngine4` (added the `IBuildEngine2`/`3`/`4` members plus an in-memory registered-task-object store). Since `IBuildEngine4` derives from `IBuildEngine`, all existing `MockEngine` usages stay source-compatible; `LocateRepositoryTests` now uses `MockEngine`. Full build clean and all `Microsoft.Build.Tasks.Git.UnitTests` (net11.0) pass.

src/Microsoft.Build.Tasks.Git.UnitTests/LocateRepositoryTests.cs
```diff
@@ -0,0 +1,86 @@
...
+        public void RelativePath_IsRejected_IndependentOfProcessCwd()
+        {
+            using var temp = new TempRoot();
+            var (repoDir, _) = CreateMinimalRepository(temp);
+
+            var originalCurrentDirectory = Directory.GetCurrentDirectory();
+            try
+            {
+                // A CWD-dependent (pre-migration) implementation would resolve "." against this repository.
+                Directory.SetCurrentDirectory(repoDir.Path);
```
tmat: I don't think we need to set CWD here. It is sufficient that we validate that the argument is checked to be an absolute path.

jankratochvilcz: https://github.com/jankratochvilcz/sourcelink/pull/1 - will make a tiny PR as a follow up if that's OK, would prefer to get this merged to unblock the rest of the migrations 🙏

tmat (approved):
OK, to clean up tests in a follow up

ViktorHofer (approved):

ViktorHofer (approved):

jankratochvilcz:
(I see Copilot squeezed in the simplification in the end before the merge)

src/Microsoft.Build.Tasks.Git.UnitTests/AssemblyInfo.cs
```diff
@@ -0,0 +1,10 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the License.txt file in the project root for more information.
+
+using Xunit;
+
+// Some tests in this assembly temporarily change the process current working directory to validate that
+// repository discovery resolves paths against the task's project directory rather than the process CWD.
+// Disable test parallelization in this assembly so that the CWD mutation can't race with other tests.
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
```
tmat, line 10: This shouldn't be needed anymore.

PR #1737, opened 2026-07-08
Simplify LocateRepository relative-path test (follow-up to #1734)

jankratochvilcz:
Follow-up to #1734 addressing review comment https://github.com/dotnet/sourcelink/pull/1734/changes#r3537777727.

The reviewer noted we don't need to set the process CWD in the relative-path test — it is sufficient to validate that a non-fully-qualified argument is rejected.

This change (`src/Microsoft.Build.Tasks.Git.UnitTests/LocateRepositoryTests.cs`):
- Removes the `Directory.SetCurrentDirectory` manipulation (and the try/finally restore) from the relative-path test.
- Renames `RelativePath_IsRejected_IndependentOfProcessCwd` to `RelativePath_IsRejected`; it now just passes a relative path (`"."`) and asserts the missing-repository warning.
- Drops the now-unused `System.IO` using.

> [!NOTE]
> This PR is stacked on #1734. Until #1734 merges, the diff here also shows #1734's commits; it will narrow to just the test simplification once #1734 is in `main`.

Verified: `RelativePath_IsRejected` and `AbsolutePath_LocatesRepository` pass on net11.0.


jankratochvilcz:
Closing as redundant. #1734 merged with the test simplification (commit 8e9e319) as its final commit, so this change is already in `main`. The review comment https://github.com/dotnet/sourcelink/pull/1734/changes#r3537777727 is fully addressed there. The diff here only appeared large because it was computed against the pre-#1734 merge-base.

